PR #9, opened 2022-10-03
Update requirements.txt to use HTTPS over SSH

mrseeker:
SSH does not work with clients that are not connected using a gitlab account.


abhi-mosaic (approved):
LGTM! Thank you for the suggestion :)

PR #11, opened 2022-10-10
Tie up weights of embedding layer and lm_head

ananyahjha93:
This has been a common practice in GPT/Bloom/OPT models according to the paper: https://aclanthology.org/E17-2025.pdf

The code for tying up weights exists within modeling_utils.py in transformers repository (https://github.com/huggingface/transformers/blob/main/src/transformers/modeling_utils.py#L1198) and is achieved by calling ``self.post_init()`` in GPT2/Bloom models.

However, I could not find the code for tying up the weights in GPT class within this repo.
On further inspection (by running ``torch.sum(model.lm_head.weight - model.transformer.wte.weight)``), I found that this is not equal to 0, which it should be if the weights were tied.

### DIFF
--- a/llm/llm/gpt.py
+++ b/llm/llm/gpt.py
@@ -118,6 +118,10 @@ class GPT(nn.Module):
                                  bias=False,
                                  device=device)
 
+        # lm_head and embedding weights should be tied: https://aclanthology.org/E17-2025.pdf
+        # HF transformers has this in transformers/modeling_utils.py; function _tie_or_clone_weights()
+        self.lm_head.weight = self.transformer.wte.weight
+
         if device != 'meta':
             self.apply(self.param_init_fn)
 


abhi-mosaic:
I think getting shared parameters to work with FSDP will require a bit of care (https://pytorch.org/docs/stable/fsdp.html), I'll try to make a parallel PR to this one and test it out, and merge/close this issue once completed. 

Thanks @ananyahjha93 for the great catch and suggestion!

dblalock:
Hi @ananyahjha93,

I believe we've [implemented this](https://github.com/mosaicml/benchmarks/blob/722be13c452bfe7b5437eec53979e502e761dd84/llm/src/mosaic_gpt.py#L150-L153) now based on your suggestion:
```python
self.transformer._fsdp_wrap = False
self.transformer.wte._fsdp_wrap = False
self.lm_head._fsdp_wrap = False
self.lm_head.weight = self.transformer.wte.weight
```

Thanks for pointing this out! Going to close but feel free to re-open if this doesn't do what you want.

PR #21, opened 2022-11-08
patch device import

dakinggg:


jacobfulano:
This is a short term fix so that GLU can work with both `v0.11.0` and `dev` branches. Ideally this will be removed in future versions

dakinggg:
Just to clarify, GLU is still broken on `v0.11.0`. This is to allow the general script (without GLU) to run on both `v0.11.0` and `dev`, because `devices` module has moved on current `dev` to a different import location.

PR #24, opened 2022-11-14
Bert pre-training and fine-tuning on GLUE

alextrott16:
Adds benchmark examples for BERT pre-training and fine-tuning on GLUE, and features support for HF models as well as a Mosaic BERT which is implemented and introduced here.

See `README.md` for a detailed description of these components.

TODO:

- [x] README: Polish (Motivation, Contents, Prerequisites, wording)
- [x] README: GLUE instructions
- [x] YAMLs: Clean up
- [x] Update to 0.11.1
- [x] Add GLUE memory fix
- [x] Improve `from_pretrained` method in Mosaic BERT
- [x] MCLI example

TODO in a future PR:
-  README: Add final speedup results
-  YAMLs: Add configuration files for experiments shown in results
-  Tests


bert/glue.py
```diff
@@ -0,0 +1,391 @@
...
+    for task_name, task_config in main_config.tasks.items():
+        if task_name not in tasks_to_run:
+            continue
+        for task_seed in task_config.get('seeds', [main_config.default_seed]):
+            run_name = f'{main_config.base_run_name}_task={task_name}_seed={str(task_seed)}'
+            logger_configs = copy.deepcopy(main_config.get('loggers', {}))
+            for logger_name, logger_config in logger_configs.items():
+                if logger_name == 'wandb':
+                    logger_config['group'] = main_config.base_run_name
+                    logger_config['name'] = run_name
```
jacobfulano, line 152: Adding small suggestions here for myself.
* It could be helpful to add the following line for plotting and filtering purposes in w&b:
```logger_config['init_kwargs'] = {'config': {'task': task_name, 'seed': task_seed}}```
* It could be helpful to allow use to define `group` in the config; if user doesn't specify, then we can set it automatically to the base_run_name

File: bert/src/glue/finetuning_jobs.py [hunk 2]
```diff
@@ -0,0 +1,749 @@
...
+
+import torch
+from torch.utils.data import DataLoader
+
+from composer.core import Callback
+from composer.core.evaluator import Evaluator
+from composer.core.types import Dataset
+from composer.loggers import LoggerDestination
+from composer.optim import ComposerScheduler, DecoupledAdamW
+try:
```
jacobfulano, line 16: Adding this comment for myself:

This try/except can be removed once composer/v0.11.1 is released and this entrypoint is tied to that release.

The new code can simply be:
```from composer.trainer.devices import Device, DeviceGPU```

File: bert/src/mosaic_bert.py [hunk 3]
```diff
@@ -0,0 +1,246 @@
...
+        pretrained_model_name = 'bert-base-uncased'
+
+    config = transformers.AutoConfig.from_pretrained(pretrained_model_name, **model_config)
+    config.return_dict = False
+    # Padding for divisibility by 8
+    if config.vocab_size % 8 != 0:
+        config.vocab_size += 8 - (config.vocab_size % 8)
+
+    if pretrained_checkpoint is not None:
+        model = BertForMaskedLM.from_pretrained(pretrained_checkpoint=pretrained_checkpoint, config=config)
```
dakinggg, line 97: Is the `pretrained_checkpoint` here a composer checkpoint? If so, it won't work because the keys will have the extra model prefix right? I could definitely be missing something here.

dakinggg (commented):
Left a handful of comments, but nothing looks amiss as far as I can tell

dblalock:
Just some high-level comments since this is a 4700-line PR:

- Maybe the answer's no, but is there a way for this to *not* be a 4700-line PR? Asking since the code is more hackable when users can easily sit down and read it all.
- Aren't we not open-sourcing Triton stuff?
- Can we have an example MCLI YAML [like in the ResNet-50 PR](https://github.com/mosaicml/benchmarks/pull/25/files#diff-7ca30a8f121213e206854b723cd0971cb991d739b1e6dd017b17830ba66da41e)? Also I like how that PR handles testing, or at least what it has so far.

bert/README.md
```diff
@@ -0,0 +1,185 @@
...
+* [Pre-train or Annotate? Domain Adaptation with a Constrained Budget](https://arxiv.org/abs/2109.04711)
+* [Foundation Models of Scientific Knowledge for Chemistry: Opportunities, Challenges and Lessons Learned](https://openreview.net/forum?id=SLX-I2MHUZ9)
+* [Domain-Specific Language Model Pretraining for Biomedical Natural Language Processing](https://arxiv.org/abs/2007.15779)
+* [LinkBERT: Pretraining Language Models with Document Links](https://arxiv.org/abs/2203.15827)
+
+One may wonder, why start with random initializations? Granted that it is better to train on domain-specific data, can't that happen as "domain adaptation" from a pretrained checkpoint? There are two reasons not to do this, one theoretical and one practical. The theory says that, because we are doing non-convex optimization, domain adaptation "may not be able to completely undo suboptimal initialization from the general-domain language model" [Gu et al., 2020](https://arxiv.org/abs/2007.15779).
+
+The practical reason is that certain outcomes are only available if the model and tokenizer are pre-trained from scratch.
```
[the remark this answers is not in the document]
samhavens: Good call

bert/main.py
```diff
@@ -0,0 +1,200 @@
...
+            pretrained_model_name=cfg.pretrained_model_name,
+            use_pretrained=cfg.get('use_pretrained', None),
+            model_config=cfg.get('model_config', None),
+            tokenizer_name=cfg.get('tokenizer_name', None),
+            gradient_checkpointing=cfg.get('gradient_checkpointing', None)
+        )
+    elif cfg.name == 'mosaic_bert':
+        return create_mosaic_bert_mlm(
+            pretrained_model_name=cfg.pretrained_model_name,
+            pretrained_checkpoint=cfg.get('pretrained_checkpoint', None),
```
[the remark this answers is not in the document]
dakinggg: oh woops, sorry, I put that on comment on the wrong line. Was supposed to be above on `use_pretrained=cfg.get('use_pretrained', None)`

bert/src/bert_layers.py
```diff
@@ -0,0 +1,855 @@
...
+            attention_probs = nn.functional.softmax(attention_scores, dim=-1)
+            attention_probs = self.dropout(attention_probs)
+            attention = torch.matmul(attention_probs, v).permute(0, 2, 1, 3)  # b s h d
+        else:
+            # Triton implementation only supports 0 attention dropout
+            convert_dtype = qkv.dtype not in [torch.float16, torch.bfloat16]
+            if convert_dtype:
+                # Triton implementation only supports fp16 and bf16
+                orig_dtype = qkv.dtype
+                qkv = qkv.to(torch.float16)
```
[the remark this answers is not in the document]
alextrott16: This will be triggered if the precision isn't fp16 or bf16. I think the assumption is that if it's not one of those two, then the original dtype is fp32, hence using fp16 for downcasting.

bert/src/mosaic_bert.py
alextrott16: No, not a composer checkpoint. This kind of HF API loads in a saved torch state_dict, so `pretrained_checkpoint` is the path to that save file. The format of the state_dict (i.e. the prefixes) would depend on how the user created the save file. We could anticipate the likely use case that they would save from the wrapped `ComposerModel`, meaning it would have the extra `model` prefix. So maybe the `from_pretrained` method should remove any such leading prefix.

alextrott16:
@dblalock Please let me know if the following addresses your comments or if you think there's more to discuss:

> Maybe the answer's no, but is there a way for this to not be a 4700-line PR? Asking since the code is more hackable when users can easily sit down and read it all.

Sigh. I don't see an easy way to trim this down, if that's what you mean. The bulkiness of the code comes from having both pre-training and fine-tuning examples (especially since GLUE requires some pretty intense scripting). We could remove the `convert_c4` stuff (since it is redundant with the LLM section) and simply refer the reader to the LLM section for instructions on preparing a local C4 copy. Please let me know if you think that would make this more digestible.

> Aren't we not open-sourcing Triton stuff?

The decision was made to showcase the full Mosaic BERT, including all relevant speedup methods. We have the green light from perf.

> Can we have an example MCLI YAML like in the ResNet-50 PR? Also I like how that PR handles testing, or at least what it has so far.

Will do! Thanks for the reference example.

bert/glue.py
```diff
@@ -0,0 +1,392 @@
...
+    num_gpus = torch.cuda.device_count()
+    results = []
+
+    with mp.Manager() as manager:
+        # workers get gpu ids from this queue
+        # to set the GPU to run on
+        gpu_queue = _setup_gpu_queue(num_gpus, manager)
+
+        ctx = mp.get_context('spawn')
+        with Pool(max_workers=min(num_gpus, len(configs)), mp_context=ctx) as pool:
```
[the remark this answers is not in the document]
alextrott16: Good idea. Will do.

bert/src/bert_layers.py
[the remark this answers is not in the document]
samhavens: In Tri's implementation it forced it into fp16, so I assumed that was better. Did not test though.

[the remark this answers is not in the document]
alextrott16: 👍  Let's keep as is

bert/yamls/glue/config.yaml
```diff
@@ -0,0 +1,81 @@
...
+# Base model config
+model:
+  name: huggingface_auto_for_sequence_classification
+  pretrained_model_name: bert-base-uncased
+
+# Tokenizer for dataset creation
+tokenizer_name: bert-base-uncased
+
+# Saving and loading
+starting_checkpoint_load_path: s3://mosaicml-internal-checkpoints-shared/alex/bert/dc-2022-07-v2/bert-dc-2022-07-baseline-alibi-glu--seed1654/checkpoints/ep0-ba68796-rank0-43383.486282wct
```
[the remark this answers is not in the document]
alextrott16: I don't think we have a way to host public checkpoints at the moment, so I'm leaving fields like this blank with comments to instruct the user to fill them in.

bert/README.md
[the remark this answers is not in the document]
alextrott16: I added some text for this. Please take a look and/or edit as you think best.

bert/README.md
```diff
@@ -0,0 +1,185 @@
...
+
+So, for example, if you want to use ALiBi positional embeddings (and [you probably should](https://ofir.io/The-Use-Case-for-Relative-Position-Embeddings/), they seem to improve LM perplexity, downstream accuracy, and allow the model to generalize to longer sequences than seen at train time), you need to train from scratch (or fine-tune from a checkpoint which was pretrained with ALiBi positional embeddings, which we will be releasing!). Or if you want to use Gated Linear Units in place of regular feedforwards ([as recommended by Noam Shazeer](https://arxiv.org/abs/2002.05202), one of the authors of the original Transformers paper), again, you have to train with them from scratch.
+
+Another good example is domain-specific tokenization. In the biomedical domain, words may be split by the pretrained BERT tokenizer in ways that make downstream tasks more difficult and computationally expensive. For example, the common drug "naloxone" in tokenized by `bert-base-uncased` tokenizer into the 4 tokens `([na, ##lo, ##xon, ##e]` [Gu et al., 2020](https://arxiv.org/abs/2007.15779), making tasks like NER more difficult and using more of the limited sequence length available.
+
+## Contents
```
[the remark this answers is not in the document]
alextrott16: GLUE is now discussed throughout. Please let me know if you consider anything to still be missing.

bert/README.md
```diff
@@ -0,0 +1,233 @@
+# Welcome!
+
+This benchmark covers pre-training and fine-tuning a BERT model from scratch. With this starter code, you'll be able to do Masked Language Modeling (MLM) pre-training on the C4 dataset and classification fine-tuning on the 8 different GLUE benchmark tasks. We also provide the source code and recipe behind our Mosaic BERT model, which you can train yourself using this repo.
```
dakinggg, line 3: "8 different GLUE benchmark tasks", should probably note that we exclude WNLI

bert/README.md
```diff
@@ -0,0 +1,233 @@
...
+## Contents
+
+You'll find in this folder:
+
+* This `README.md`
+* `main.py` — A straightforward script for parsing YAMLs, building a [Composer](https://github.com/mosaicml/composer) Trainer, and kicking off an MLM pre-training training job, locally or on Mosaic's cloud
+* `glue.py` - A more complex script for parsing YAMLs and orchestrating the numerous fine-tuning training jobs across the 8 GLUE tasks, locally or on Mosaic's cloud
+* `convert_c4.py` — Code for creating a streaming C4 dataset, which can be used for pre-training. See [Dataset preparation](#Dataset-preparation)
+* `src/data_c4.py` — A [MosaicML streaming dataset](https://docs.mosaicml.com/projects/streaming/en/latest/) that can be used with a vanilla PyTorch dataloader or [Composer](https://github.com/mosaicml/composer)
+* `src/hf_bert.py` — Code for creating a HuggingFace BERT model for MLM (pre-training) or classification (GLUE fine-tuning), wrapped in [`ComposerModel`](https://docs.mosaicml.com/en/v0.11.1/composer_model.html), for compatibility with the [Composer Trainer](https://docs.mosaicml.com/en/v0.11.1/api_reference/generated/composer.Trainer.html#composer.Trainer)
```
dakinggg, line 14: Should this link to HuggingFaceModel instead of ComposerModel? The wrapper is technically a HuggingFaceModel (which is a ComposerModel).

bert/README.md
```diff
@@ -0,0 +1,233 @@
...
+* `src/mosaic_bert.py` — Code for creating a Mosaic BERT model for MLM (pre-training) or classification (GLUE fine-tuning), wrapped in [`ComposerModel`](https://docs.mosaicml.com/en/v0.11.1/composer_model.html), for compatibility with the [Composer Trainer](https://docs.mosaicml.com/en/v0.11.1/api_reference/generated/composer.Trainer.html#composer.Trainer)
+* `src/bert_padding.py` — Source code for Mosaic BERT that supports reshaping tensors in order avoid inefficiencies due to padding
+* `src/bert_layers.py` — Source code for Mosaic BERT that re-implements BERT layers with our custom speed up methods built it, with an eye towards HuggingFace API compatibility
```
dakinggg, line 17: "built it" -> "built into it"

bert/README.md
```diff
@@ -0,0 +1,233 @@
...
+composer main.py yamls/mosaic-bert-base-uncased.yaml
+```
+
+**Note:** The `yamls/*.yaml` files are intended to be used with `main.py`.
+
+**Please remember** to modify the reference YAMLs (e.g., `yamls/mosiac-bert-base-uncased.yaml`) to customize saving and loading locations -- only the YAMLs in `yamls/test-cpu/` are ready to use out-of-the-box.
+
+## GLUE fine-tuning
+
+The GLUE benchmark measures the average performance across 8 NLP classification tasks. This performance is typically used to evaluate the quality of the pre-training: once you have a set of weights from your MLM task, you fine-tune those weights separately for each task and average the results, with higher averages indicating higher pre-training quality.
```
dakinggg, line 153: again, mention excluding WNLI

bert/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+To handle this complicated fine-tuning pipeline, we provide the `glue.py` script, which supports parallelizing each of these fine-tuning jobs across all the GPUs on your node.
```
dakinggg, line 155: suggest adding: `Note: because of the small dataset/batch size of the GLUE tasks, this script parallelizes tasks across individual GPUs, rather than parallelizing each task across multiple GPUs` or something like that.

bert/README.md
```diff
@@ -0,0 +1,233 @@
...
+
+Once you have modified the YAMLs in `yamls/glue/` to reference your pre-trained checkpoint as the GLUE starting point, run the `glue.py` script using the standard `python` launcher (we don't use the `composer` launcher here because `glue.py` does its own multi-process orchestration):
+
+```bash
+composer glue.py yamls/glue/mosaic-bert-base-uncased.yaml
```
dakinggg, line 160: should be `python glue.py`

bert/README.md
```diff
@@ -0,0 +1,233 @@
...
+```
+
+Aggregate GLUE scores will be printed out at the end of the script and can also be tracked using Weights and Biases, if enabled via the YAML.
```
dakinggg, line 163: suggest "using Weights and Biases" -> "using Weights and Biases (any of the other (composer supported loggers)[https://docs.mosaicml.com/en/latest/trainer/logging.html#available-loggers] can be added easily as well)"

bert/yamls/test-cpu/glue.yaml
```diff
@@ -0,0 +1,56 @@
+# Use this YAML to verify that GLUE fine-tuning works. Runs on CPU.
+# From `benchmarks/bert`, run: 
+#   `python main.py yamls/test-cpu/glue.yaml` to run using the HuggingFace BERT
```
dakinggg, line 3: these commands should be `glue.py` I believe?

bert/README.md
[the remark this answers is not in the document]
dakinggg: :+1: rereviewed the readme

bert/README.md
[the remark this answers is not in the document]
dakinggg: LGTM

bert/README.md
alextrott16: Oops. Was supposed to be "built in".

bert/README.md
alextrott16: Fixing this here by just removing detail (since it's an intro paragraph). The WNLI exclusion detail will be mentioned below.

bert/src/bert_layers.py
```diff
@@ -0,0 +1,859 @@
...
+
+        # self.LayerNorm is not snake-cased to stick with TensorFlow model variable name and be able to load
+        # any TensorFlow checkpoint file
+        self.LayerNorm = nn.LayerNorm(config.hidden_size, eps=config.layer_norm_eps)
+        self.dropout = nn.Dropout(config.hidden_dropout_prob)
+        self.register_buffer(
+            "token_type_ids", torch.zeros(config.max_position_embeddings, dtype=torch.long), persistent=False
+        )
+
+    def forward(
```
jacobfulano, line 50: Do we want more extensive documentation here?

bert/src/bert_layers.py
```diff
@@ -0,0 +1,859 @@
+# coding=utf-8
```
jacobfulano, line 1: It would be helpful somewhere (e.g. in a README) to explain in detail what each architecture choice such as GLU and ALiBi does, and how it is implemented in the code. Not blocking.

bert/src/bert_padding.py
```diff
@@ -0,0 +1,138 @@
+
```
jacobfulano, line 1: This is honestly pretty confusing. More documentation would be helpful at some point. Not a blocker.

bert/src/flash_attn_triton.py
```diff
@@ -0,0 +1,820 @@
+"""
```
jacobfulano, line 1: This is really confusing to read. Glad it works though!

bert/src/mosaic_bert.py
```diff
@@ -0,0 +1,246 @@
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Implements a BERT wrapper around a :class:`.ComposerTransformer`."""
```
jacobfulano, line 4: Maybe "Mosaic BERT" instead of "BERT" just to be clear?

bert/yamls/glue/mcloud_run.yaml
```diff
@@ -0,0 +1,105 @@
...
+      trainer_kwargs:
+        save_num_checkpoints_to_keep: 0
+    mrpc:
+      seeds: [19, 8364, 717, 10536, 90166]
+      trainer_kwargs:
+        save_num_checkpoints_to_keep: 0
+    cola:
+      seeds: [19, 8364, 717, 10536]
+      trainer_kwargs:
+        save_num_checkpoints_to_keep: 0
```
jacobfulano, line 105: You don't want to include your slightly modified GLUE parameters?

```
    mrpc:
      seeds: [19, 8364, 717, 10536, 90166]
      trainer_kwargs:
        save_num_checkpoints_to_keep: 0
    cola:
      seeds: [19, 8364, 717, 10536]
      trainer_kwargs:
        save_num_checkpoints_to_keep: 0
        max_duration: 40ep
        batch_size: 128
   ```

File: bert/glue.py (hunk 1 above)
Comment (jacobfulano) [reply bert/glue.py:151]: I'm going to hold off on this for now until I can verify that it works for my own pipeline

**Review by jacobfulano — APPROVED** [review]
This PR is a welcome addition to the benchmarks repo. While it could use some more documentation (especially in the code used for Mosaic BERT), I can verify that it runs successfully across many scenarios. BERT is dead; long live BERT!

File: bert/yamls/glue/mcloud_run.yaml (hunk 23 above)
Comment (jacobfulano) [reply bert/yamls/glue/mcloud_run.yaml:105]: (these are the updated details/hyperparameters I have been using in my GLUE sweeps)

Comment (alextrott16) [reply bert/yamls/glue/mcloud_run.yaml:105]: It is looking less likely that we need the specific HPs, so we can leave it out for now. Also, we will update the recipes once the results are finalized.

File: bert/src/bert_layers.py (hunk 18 above)
Comment (alextrott16) [reply bert/src/bert_layers.py:86]: Good suggestion. Added more exposition throughout.

File: bert/src/bert_layers.py (hunk 19 above)
Comment (alextrott16) [reply bert/src/bert_layers.py:1]: Added a header and pointed out which modules handle which new choices throughout.

File: bert/src/bert_padding.py (hunk 20 above)
Comment (alextrott16) [reply bert/src/bert_padding.py:1]: I think we'll leave this and the flash attention documentation for now, since it's pretty low-level code that the user will hopefully not need to understand. If that's leading to confusion, though, we'll revisit.

File: bert/README.md [hunk 24]
```diff
@@ -0,0 +1,244 @@
...
+
+Another good example is domain-specific tokenization. In the biomedical domain, words may be split by the pre-trained BERT tokenizer in ways that make downstream tasks more difficult and computationally expensive. For example, the common drug "naloxone" in tokenized by `bert-base-uncased` tokenizer into the 4 tokens `[na, ##lo, ##xon, ##e]` [(Gu et al., 2020)](https://arxiv.org/abs/2007.15779), making tasks like NER more difficult and using more of the limited sequence length available.
+
+Now that we've convinced you that you should train a Mosaic BERT from scratch, let's get into the how :) 
+
+# Setup
+
+Here's what you need to get started with our BERT implementation:
```
dblalock, line 68: It takes several screens of scrolling to get to this. Can we start the readme with:

- A couple intro sentences
- Installation / dependencies
- Quick start (a command or two that should "just work," even if only on synthetic data or something)

Want to minimize the barrier to getting started and deliver value to users ASAP.

bert/README.md
```diff
@@ -0,0 +1,244 @@
...
+
+* Use a Docker image with PyTorch 1.12+, e.g. [MosaicML's PyTorch base image](https://hub.docker.com/r/mosaicml/pytorch/tags)
+  * Recommended tag: `mosaicml/pytorch:1.12.1_cu116-python3.9-ubuntu20.04`
+  * This image comes pre-configured with the following dependencies:
+    * PyTorch Version: 1.12.1
+    * CUDA Version: 11.6
+    * Python Version: 3.9
+    * Ubuntu Version: 20.04
```
dblalock, line 76: Can we turn this into just `docker pull mosaicml/pytorch:1.12.1_cu116-python3.9-ubuntu20.04` + a link to more info about what's in this image if you're curious (maybe at the bottom of this readme, but ideally shared with other benchmarks since they all have this).

Trying to get this section down to a few lines the user can copy-paste to get started. Something like:

```
git clone https://github.com/mosaicml/benchmarks.git
cd bert
docker pull mosaicml/pytorch:1.12.1_cu116-python3.9-ubuntu20.04
...
```

Alternatively, we could have something like this in a Quickstart section at the top and this could be the more detailed version we link to.

bert/README.md
```diff
@@ -0,0 +1,244 @@
...
+* Use a system with NVIDIA GPU(s)
+
+* Install requirements via: `pip install -r requirements.txt`
+
+* Prepare a local copy of the C4 dataset via instructions below.
+
+## Dataset preparation
+
+To run pre-training, you'll need to make yourself a local copy of the C4 pre-training dataset.
+If you only want to profile these BERTs, we recommend that you **only download and prepare the `val` split**,
```
dblalock, line 87:
```suggestion
If you only want to profile the speed of these BERTs, we recommend that you **only download and prepare the `val` split**,
```
Not confident that people with zero perf background will get that "profile" always refers to timing, as opposed to just being a synonym for "characterize" or "analyze".

dblalock: Also, is this a common enough case to merit the space it takes to mention it?

bert/README.md
```diff
@@ -0,0 +1,244 @@
...
+and use it for both train and eval in your script. Just change `split: train` to `split: val` in your run YAML.
+Alternatively, feel free to substitute our dataloader with one of your own in the script [main.py](./main.py#L101)!
+
+In this benchmark, we train BERTs on the [C4: Colossal, Cleaned, Common Crawl dataset](https://huggingface.co/datasets/c4).
```
dblalock, line 91: should define C4 the first time we use it, not here

bert/README.md
```diff
@@ -0,0 +1,244 @@
...
+We first convert the dataset from its native format (a collection of zipped JSONs)
+to MosaicML's streaming dataset format (a collection of binary `.mds` files).
+Once in `.mds` format, we can store the dataset in a central location (filesystem, S3, GCS, etc.)
+and stream the data to any compute cluster, with any number of devices, and any number of CPU workers, and it all ~ just works ~ .
```
dblalock, line 95:
```suggestion
and stream the data to any compute cluster, with any number of devices, and any number of CPU workers, and it all ~just works~ .
```
Assuming the lack of whitespace doesn't mess with the markdown. Otherwise just rm the tildes.

bert/README.md
```diff
@@ -0,0 +1,244 @@
...
+
+If you have configured a compute cluster to work with the MosaicML Cloud, you can use the `mcloud_run.yaml` reference YAMLs for examples of how to run pre-training and fine-tuning remotely!
+
+Once you have filled in the missing YAML fields (and made any other modifications you want), you can launch pre-training by simply running:
+
+```bash
+mcli run -f yamls/mcloud_run.yaml
+```
+
+Similarly, for GLUE fine-tuning just fill in the missing YAML fields (e.g., to use the pre-training checkpoint as the starting point) and run:
```
dblalock, line 186: Also, can we enumerate which fields are missing + split this into multiple sentences?

```suggestion
Similarly, for GLUE fine-tuning, just fill in the missing YAML fields (e.g., to use the pre-training checkpoint as the starting point) and run:
```

bert/README.md
```diff
@@ -0,0 +1,244 @@
...
+```
+
+You should see logs being printed to your terminal.
+You can also easily enable other experiment trackers like Weights and Biases or CometML by using [Composer's logging integrations](https://docs.mosaicml.com/en/v0.11.1/trainer/logging.html).
+
+# Contact Us
+
+If you run into any problems with the code, please file Github issues directly to this repo.
+
+If you want train BERT-style models on MosaicML Cloud, reach out to us at [demo@mosaicml.com](mailto:demo@mosaicml.com)!
```
dblalock, line 244:
```suggestion
If you want to train BERT-style models on MosaicML Cloud, reach out to us at [demo@mosaicml.com](mailto:demo@mosaicml.com)!
```

bert/README.md
```diff
@@ -0,0 +1,244 @@
...
+mcli run -f yamls/glue/mcloud_run.yaml
+```
+
+## Multi-node training
+
+To train with high performance on *multi-node* clusters, the easiest way is with MosaicML Cloud ;)
+
+But if you want to try this manually on your own cluster, then just provide a few variables to `composer`, either directly via CLI or via environment variables. Then launch the appropriate command on each node.
+
+**Note:** multi-node training will only work `main.py` (the `glue.py` script handles its own orchestration across devices and is not built to be used with the `composer` launcher).
```
dblalock, line 198:
```suggestion
**Note:** multi-node training will only work with `main.py`; the `glue.py` script handles its own orchestration across devices and is not built to be used with the `composer` launcher.
```

bert/yamls/glue/mcloud_run.yaml
```diff
@@ -0,0 +1,105 @@
...
+integrations:
+  - integration_type: "git_repo"
+    git_repo: mosaicml/benchmarks
+command: |
+  cd benchmarks/bert
+  pip install -r requirements.txt
+  composer main.py /mnt/config/parameters.yaml
+
+# `parameters` copied from `glue/mosaic-bert-base-uncased.yaml`
+parameters:
```
dblalock, line 25: Would still be great to not duplicate all the params in these files. Shouldn't it work without this by just having the command point to the regular yaml?

bert/convert_c4.py
```diff
@@ -0,0 +1,128 @@
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""C4 streaming dataset conversion scripts."""
```
dblalock, line 4:
```suggestion
"""C4 streaming dataset conversion script."""
```

dblalock (approved):
Main thing is I went through the README in quite a bit of detail since this is the most user-facing aspect here. Summary of my high-level comments:

- The MCLI YAMLs really need to indirect to the regular yamls rather than replicating their contents, although not necessarily in this PR
- The README needs to get to the part where the user installs and runs stuff way faster. Like, they should see commands to paste as soon as they scroll down to the README.
- We should put in a little content reassuring the reader that you don't have to use YAML and there isn't some scary configuration language going on here
- In general, the prose feels like it's trying to be short as measured by periods, not as measured by actual length, which yields really long sentences that you keep thinking will end (often with a lot of parentheses), but then they don't, and the result is hard to read

I do want to get this merged asap though so that I can finish off the linting + local testing infra. There's going to be a lot of pyright fixes in a 5k line PR...

bert/yamls/glue/mcloud_run.yaml
alextrott16: Yes, it would work if we instead did `composer main.py yamls/main/___.yaml` in the command, but that would require  having to commit/push changes to `yamls/main/___.yaml` before running it using `mcli`. I personally find that very tedious, which is why I have it set up this way.

dblalock: ah, so I think the current approach is screwed whether we do this or not. Right now these mcli yamls *only* work if you're both:

1. checking in all your changes
2. pushing your code to `mosaicml/benchmarks` 

What we actually need to do is switch to using an MCLI local integration, which will "just work" no matter where the user's code is / what files they change.

I'm planning on switching everything to local integrations in the next couple days unless some NLP person has bandwidth to take this on (in theory should take like 10min).

dblalock: I'm considering switching to a local integration out of scope for this PR though.

bert/README.md
```diff
@@ -131,59 +136,93 @@ python src/data_c4.py s3://my-bucket/my-copy-c4 /tmp/cache-c4
...
 
 **Also remember** that if you only downloaded the `val` split, you need to make sure your train_dataloader is pointed at that split.
 Just change `split: train` to `split: val` in your YAML.
-This is already done in the testing YAML `yamls/test-cpu/main.py`, which you can also use to test your configuration.
+This is already done in the testing YAML `yamls/test/main.py`, which you can also use to test your configuration (see [Quick Start](#quick-start)).
 
 ## MLM pre-training
 
-We run the `main.py` script using our `composer` launcher, which generates N processes (1 process per GPU device).
+To get the most out of your pre-training budget, we recommend using **Mosiac BERT**! You can read more [below](#mosaic-bert).
```
dblalock, line 147:
```suggestion
To get the most out of your pre-training budget, we recommend using **Mosaic BERT**! You can read more [below](#mosaic-bert).
```

bert/README.md
```diff
@@ -131,59 +136,93 @@ python src/data_c4.py s3://my-bucket/my-copy-c4 /tmp/cache-c4
...
+# It takes about 11.5 hours on a single node with 8 A100_80g GPUs.
+composer main.py yamls/main/hf-bert-base-uncased.yaml
 
-**Note:** The `yamls/*.yaml` files are intended to be used with `main.py`.
+# This will pre-train a Mosaic BERT that reaches the same downstream accuracy in roughly 1/3 the time.
+composer main.py yamls/main/mosaic-bert-base-uncased.yaml
+```
 
-**Please remember** to modify the reference YAMLs (e.g., `yamls/mosiac-bert-base-uncased.yaml`) to customize saving and loading locations -- only the YAMLs in `yamls/test-cpu/` are ready to use out-of-the-box.
+**Please remember** to modify the reference YAMLs (e.g., `yamls/main/mosiac-bert-base-uncased.yaml`) to customize saving and loading locations—-only the YAMLs in `yamls/test/` are ready to use out-of-the-box. See the [configs](#configs) section for more detail.
```
dblalock, line 161:
```suggestion
**Please remember** to modify the reference YAMLs (e.g., `yamls/main/mosaic-bert-base-uncased.yaml`) to customize saving and loading locations—only the YAMLs in `yamls/test/` are ready to use out-of-the-box. See the [configs](#configs) section for more detail.
```
The first dash is already an em dash; having both looks weird in the rendered readme

bert/README.md
```diff
@@ -131,59 +136,93 @@ python src/data_c4.py s3://my-bucket/my-copy-c4 /tmp/cache-c4
...
+
+In other words, you're free to modify this starter code to suit your project and aren't tied to using YAMLs in your workflow.
+
+## main.py
 
-**Please remember** to modify the reference YAMLs (e.g., `yamls/glue/mosiac-bert-base-uncased.yaml`) to customize saving and loading locations -- only the YAMLs in `yamls/test-cpu/` are ready to use out-of-the-box.
+Before using the configs in `yamls/main/` when running `main.py`, you'll need to fill in:
 
-### Running on the MosaicML Cloud
+* `save_folder` - This will determine where model checkpoints are saved. Note that it can depend on `run_name`, so you should avoid re-using the same run name across multiple training runs.
```
dblalock, line 202: How can it depend on run name? Is there an example that users can paste in?

bert/README.md
```diff
@@ -224,21 +263,71 @@ composer --world_size 16 --node_rank 1 --master_addr 0.0.0.0 --master_port 7501
...
+
+So, for example, if you want to use ALiBi positional embeddings (and [you probably should](https://ofir.io/The-Use-Case-for-Relative-Position-Embeddings/), they seem to improve LM perplexity, downstream accuracy, and allow the model to generalize to longer sequences than seen at train time), you need to train from scratch (or fine-tune from a checkpoint which was pre-trained with ALiBi positional embeddings, which we will be releasing!). Or if you want to use Gated Linear Units in your feedforward layers ([as recommended by Noam Shazeer](https://arxiv.org/abs/2002.05202), one of the authors of the original Transformers paper), again, you have to train with them from scratch.
+
+Another good example is domain-specific tokenization. In the biomedical domain, words may be split by the pre-trained BERT tokenizer in ways that make downstream tasks more difficult and computationally expensive. For example, the common drug "naloxone" in tokenized by `bert-base-uncased` tokenizer into the 4 tokens `[na, ##lo, ##xon, ##e]` [(Gu et al., 2020)](https://arxiv.org/abs/2007.15779), making tasks like NER more difficult and using more of the limited sequence length available.
```
dblalock, line 320:
```suggestion
Another good example is domain-specific tokenization. In the biomedical domain, words may be split by the pre-trained BERT tokenizer in ways that make downstream tasks more difficult and computationally expensive. For example, the common drug "naloxone" is tokenized by `bert-base-uncased` tokenizer into the 4 tokens `[na, ##lo, ##xon, ##e]` [(Gu et al., 2020)](https://arxiv.org/abs/2007.15779), making tasks like NER more difficult and using more of the limited sequence length available.
```

bert/README.md
```diff
@@ -224,21 +263,71 @@ composer --world_size 16 --node_rank 1 --master_addr 0.0.0.0 --master_port 7501
...
+
+* [Downstream Datasets Make Surprisingly Good Pretraining Corpora](https://arxiv.org/abs/2209.14389)
+* [Quality Not Quantity: On the Interaction between Dataset Design and Robustness of CLIP](https://arxiv.org/abs/2208.05516)
+* [Insights into Pre-training via Simpler Synthetic Tasks](https://arxiv.org/abs/2206.10139)
+* [Pre-train or Annotate? Domain Adaptation with a Constrained Budget](https://arxiv.org/abs/2109.04711)
+* [Foundation Models of Scientific Knowledge for Chemistry: Opportunities, Challenges and Lessons Learned](https://openreview.net/forum?id=SLX-I2MHUZ9)
+* [Domain-Specific Language Model Pretraining for Biomedical Natural Language Processing](https://arxiv.org/abs/2007.15779)
+* [LinkBERT: Pretraining Language Models with Document Links](https://arxiv.org/abs/2203.15827)
+
+In addition to being able to take advantage of pre-training on in-domain data, training from scratch means that you control the data start to finish. Publicly available pre-training corpuses cannot be used in many commercial cases due to legal considerations. Our starter code can easily be modified to handle custom datasets beyond the C4 example we provide.
```
dblalock, line 312:
```suggestion
In addition to being able to take advantage of pre-training on in-domain data, training from scratch means that you control the data start to finish. Publicly available pre-training corpora cannot be used in many commercial cases due to legal considerations. Our starter code can easily be modified to handle custom datasets beyond the C4 example we provide.
```

bert/README.md
```diff
@@ -224,21 +263,71 @@ composer --world_size 16 --node_rank 1 --master_addr 0.0.0.0 --master_port 7501
...
+
+One may wonder, why start from scratch when public data isn't a concern? Granted that it is better to train on domain-specific data, can't that happen as "domain adaptation" from a pre-trained checkpoint? There are two reasons not to do this, one theoretical and one practical. The theory says that, because we are doing non-convex optimization, domain adaptation "may not be able to completely undo suboptimal initialization from the general-domain language model" [(Gu et al., 2020)](https://arxiv.org/abs/2007.15779).
```
dblalock, line 314:
```suggestion
One may wonder, why start from scratch when public data *isn't* a concern? Granted that it is better to train on domain-specific data, can't that happen as "domain adaptation" from a pre-trained checkpoint? There are two reasons not to do this, one theoretical and one practical. The theory says that, because we are doing non-convex optimization, domain adaptation "may not be able to completely undo suboptimal initialization from the general-domain language model" [(Gu et al., 2020)](https://arxiv.org/abs/2007.15779).
```

bert/README.md
```diff
@@ -224,21 +263,71 @@ composer --world_size 16 --node_rank 1 --master_addr 0.0.0.0 --master_port 7501
...
+
+The practical reason is that certain outcomes are only available if the model and tokenizer are pre-trained from scratch.
+
+So, for example, if you want to use ALiBi positional embeddings (and [you probably should](https://ofir.io/The-Use-Case-for-Relative-Position-Embeddings/), they seem to improve LM perplexity, downstream accuracy, and allow the model to generalize to longer sequences than seen at train time), you need to train from scratch (or fine-tune from a checkpoint which was pre-trained with ALiBi positional embeddings, which we will be releasing!). Or if you want to use Gated Linear Units in your feedforward layers ([as recommended by Noam Shazeer](https://arxiv.org/abs/2002.05202), one of the authors of the original Transformers paper), again, you have to train with them from scratch.
```
dblalock, line 318:
```suggestion
So, for example, if you want to use ALiBi positional embeddings (and [you probably should](https://ofir.io/The-Use-Case-for-Relative-Position-Embeddings/) since they improve LM perplexity, downstream accuracy, and generalization across sequences lengths), you need to train from scratch (or fine-tune from a checkpoint which was pre-trained with ALiBi positional embeddings, which we will be releasing!). Or if you want to use Gated Linear Units in your feedforward layers ([as recommended by Noam Shazeer](https://arxiv.org/abs/2002.05202), one of the authors of the original Transformers paper), again, you have to train with them from scratch.
```

bert/README.md
```diff
@@ -224,21 +263,71 @@ composer --world_size 16 --node_rank 1 --master_addr 0.0.0.0 --master_port 7501
...
+* [LinkBERT: Pretraining Language Models with Document Links](https://arxiv.org/abs/2203.15827)
+
+In addition to being able to take advantage of pre-training on in-domain data, training from scratch means that you control the data start to finish. Publicly available pre-training corpuses cannot be used in many commercial cases due to legal considerations. Our starter code can easily be modified to handle custom datasets beyond the C4 example we provide.
+
+One may wonder, why start from scratch when public data isn't a concern? Granted that it is better to train on domain-specific data, can't that happen as "domain adaptation" from a pre-trained checkpoint? There are two reasons not to do this, one theoretical and one practical. The theory says that, because we are doing non-convex optimization, domain adaptation "may not be able to completely undo suboptimal initialization from the general-domain language model" [(Gu et al., 2020)](https://arxiv.org/abs/2007.15779).
+
+The practical reason is that certain outcomes are only available if the model and tokenizer are pre-trained from scratch.
```
dblalock, line 316:
```suggestion
The practical reason to train from scratch is that it allows you to modify your model or tokenizer.
```

dblalock (approved):
LGTM. Had a few typo / local clarity suggestions but I think all but one can just be committed via github. Feel free to hit the big green button as soon as you're satisfied.

samhavens:
I merged all of @dblalock 's suggested README edits that were completely non-controversial spelling/grammar fixes as I couldn't imagine that requiring consensus

PR #26, opened 2022-11-18
Update LLM benchmark with eval, HF models

bmosaicml:
This PR does the following:

* Add the EleutherAI evaluation harness (forked at https://github.com/bmosaicml/lm-evaluation-harness-composergpt-integration)
* Add a HF CausalLM example
* Fix dependencies
* Fix attention mask issue for `TorchCausalAttention` when `MosaicGPT` is built with `device='meta'`
* Make `device_eval_batch_size` more explicit


llm/main.py
```diff
@@ -58,37 +58,45 @@ def build_scheduler(cfg):
...
+    device_eval_batch_size, _ = calculate_batch_size_info(
+        cfg.global_eval_batch_size, cfg.device_eval_microbatch_size
+    )
 
     cfg.n_gpus = dist.get_world_size()
     cfg.device_train_batch_size = device_train_batch_size
     cfg.device_train_grad_accum = device_train_grad_accum
     cfg.device_eval_batch_size = device_eval_batch_size
-    cfg.device_eval_microbatch_size = device_eval_microbatch_size
+    cfg.device_eval_microbatch_size = device_eval_batch_size
```
abhi-mosaic, line 94: I'm a bit worried here, I think Composer also has an OOM checker for eval, but if we want to be extra safe, we should probably not try to rely on it, and just use safe config values. For that reason I think we should have the user config include a `device_eval_batch_size` that is set to a safe value, and not worry too much about the global eval batch size, as we're not doing any optimization.

The failure case right now is if `global_eval_batch_size=128` or something is hardcoded, then I can either hit a bug if I try to run on more GPUs, or run into an OOM if I run on too few GPUs.

abhi-mosaic (commented):
Just did a first pass! I will do some direct testing soon and maybe try to think about the `eval_batch_size` config

llm/main.py
abhi-mosaic: Fixed this in a way that definitely won't break, but is a bit conservative (basically, not relying on Composer's auto eval loop)

llm/yamls/mosaic_gpt/760m.yaml
```diff
@@ -82,6 +89,7 @@ grad_clip_norm: 1.0
 
 # System
 seed: 17
+device_eval_batch_size: 4
 device_train_microbatch_size: 4
```
vchiley, line 93: If the 1b model support `device batch_size = 16`, then 760m model should also be set to at least 16.

llm/yamls/mosaic_gpt/350m.yaml
```diff
@@ -82,6 +89,7 @@ grad_clip_norm: 1.0
 
 # System
 seed: 17
+device_eval_batch_size: 8
 device_train_microbatch_size: 8
```
vchiley, line 93: If the 1b model support `device batch_size = 16`, then 350m model should also be set to at least 16.

llm/yamls/hf_causal_lm/gpt2-small.yaml
```diff
@@ -0,0 +1,117 @@
...
+max_seq_len: &max_seq_len 1024
+tokenizer_name: &tokenizer_name gpt2
+
+# Run Name
+run_name: gpt2-small
+
+# Model
+model:
+  name: hf_causal_lm
+  hf_config: gpt2
```
vchiley, line 12: To get the HF model I've also been using just `hf_config: gpt2`, but are there any other config details we should consider adding???
(This is less of a review and more of an open question which should probably be ignored)

llm/src/tokenizer.py
```diff
@@ -0,0 +1,75 @@
...
+
+    @abstractmethod
+    def __call__(self, *args, **kwargs):
+        raise NotImplementedError
+
+    @abstractmethod
+    def encode(self, x):
+        raise NotImplementedError
+
+    def decode(self, x):
```
vchiley, line 26: if `def encode(self, x):` has the `@abstractmethod` decorator, should `def decode(self, x):` also have the same decorator?

llm/src/mosaic_gpt.py
```diff
@@ -27,13 +27,26 @@ def __init__(self, cfg: Mapping[str, Any], device: str = None):
...
+        # Two important disclaimers
+        # 1. Torch uses additive attention. If your attn_mask/key_padding mask is a float tensor, it will add the floats
+        #   directly to your attention matrix. If they are boolean masks, True will be converted to -inf before adding the
+        #   mask to your attentions. See https://pytorch.org/docs/stable/generated/torch.nn.MultiheadAttention.html#torch.nn.MultiheadAttention.forward
+        #   Basically True/-inf indicates tokens we do not want to attend to.
+        #
+        # 2. This is is the exact opposite behavior of Huggingface's tokenizers, which use the convention that True denotes tokens
+        #   we do want to attend to. See https://huggingface.co/docs/transformers/glossary#attention-mask
+        #
+        return self.mha(x, x, x,
```
vchiley, line 45: do we want to note that this (and the Flash Attn) is multi headed **self** attn?
(we currently refer to it as just multi headed attn)

llm/src/evaluation/model_loading.py
```diff
@@ -0,0 +1,98 @@
...
+from src.tokenizer import TOKENIZER_REGISTRY, LLMTokenizer
+from transformers import AutoModelForCausalLM, AutoTokenizer
+
+DEVICE = torch.device('cuda') if torch.cuda.is_available() else torch.device('cpu')
+CHECKPOINT_DIR = f"{os.path.dirname(__file__)}/model_checkpoints"
+
+def get_checkpoint_name_from_path(path: str) -> str:
+    """To go from checkpoint name to path, replace | with /"""
+    return path.lstrip('/').replace('/', '|')
+
```
vchiley, line 23: \n

llm/src/evaluation/model_loading.py
```diff
@@ -0,0 +1,98 @@
...
+from composer.utils.file_helpers import get_file
+from composer.utils.object_store import S3ObjectStore
+from omegaconf import OmegaConf as om
+from src.model_registry import COMPOSER_MODEL_REGISTRY
+from src.tokenizer import TOKENIZER_REGISTRY, LLMTokenizer
+from transformers import AutoModelForCausalLM, AutoTokenizer
+
+DEVICE = torch.device('cuda') if torch.cuda.is_available() else torch.device('cpu')
+CHECKPOINT_DIR = f"{os.path.dirname(__file__)}/model_checkpoints"
+
```
vchiley, line 19: \n

llm/src/evaluation/model_loading.py
```diff
@@ -0,0 +1,98 @@
...
+    return path.lstrip('/').replace('/', '|')
+
+def download_starting_checkpoints(path_to_download: str) -> List[str]:
+    """Downloads the pretrained checkpoints to start from. Currently only supports S3 and URLs"""
+    load_object_store = None
+    parsed_first_checkpoint = urlparse(path_to_download)
+    if parsed_first_checkpoint.scheme == "s3":
+        load_object_store = S3ObjectStore(bucket=parsed_first_checkpoint.netloc)
+
+
```
vchiley, line 31: y 2 \n?

llm/src/evaluation/model_loading.py
```diff
@@ -0,0 +1,98 @@
...
+    n_params = sum(p.numel() for p in model.parameters())
+    print(f'{n_params=:.2e}')
+
+    tokenizer = TOKENIZER_REGISTRY[cfg.tokenizer.type](**cfg.tokenizer.args)
+    return {
+        "model": model.model,
+        "tokenizer": tokenizer,
+        "precision": cfg.precision
+    }
+
```
vchiley, line 94: \n

llm/src/evaluation/model_checkpoints/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
+!.gitignore
```
vchiley, line 2: \n at eof

llm/src/evaluation/eval_reports/.gitignore
```diff
@@ -0,0 +1,2 @@
+*
+!.gitignore
```
vchiley, line 2: \n at eof

llm/src/evaluation/eval.py
```diff
@@ -0,0 +1,145 @@
...
+    parser.add_argument('--tasks', metavar="task_name", type=str, nargs='+')
+    parser.add_argument('--num_fewshots', metavar="num_fewshots", type=int, nargs='+')
+    parser.add_argument('--debug', action='store_true', default=False)
+    parser.add_argument('--partial_eval_mode', action='store_true', default=False)
+
+
+    args = parser.parse_args()
+    print(args)
+    if args.debug:
+        import pdb; pdb.set_trace()
```
vchiley, line 136: pretty random place to put a pdb trace. remove

llm/src/evaluation/eval.py
```diff
@@ -0,0 +1,145 @@
...
+
+
+    """
+    parser = argparse.ArgumentParser(description='Run the EleutherAI eval harness on ComposerGPT models')
+    parser.add_argument('--experiment_name', metavar="experiment_name", type=str)
+    parser.add_argument('--model_type', metavar="model_type", type=str)
+    parser.add_argument('--model_args',  nargs='+', metavar="checkpoint_path", type=str)
+    parser.add_argument('--tasks', metavar="task_name", type=str, nargs='+')
+    parser.add_argument('--num_fewshots', metavar="num_fewshots", type=int, nargs='+')
+    parser.add_argument('--debug', action='store_true', default=False)
```
vchiley, line 129: remove

llm/src/evaluation/eval.py
```diff
@@ -0,0 +1,145 @@
...
+        Your config must also be available locally
+
+        python -m src.evaluation.eval --experiment_name gpt_neo_125m \
+            --model_type composer_checkpoint \
+            --model_args checkpoint=s3://mosaicml-internal-checkpoints-shared/jeremy/gpt-eval/gpt-neo-125m-5rbfZF/checkpoints/ep0-ba80000-rank0.pt \
+                config=yamls/mosaic_gpt/125m.yaml \
+            --tasks lambada \
+            --num_fewshot 0 1 10
+
+
```
vchiley, line 121: extra newlines

llm/src/evaluation/eval.py
```diff
@@ -0,0 +1,145 @@
...
+        fewshot_exp_results = results[num]['results']
+        for task_name in fewshot_exp_results:
+            accuracy = fewshot_exp_results[task_name].get("acc", "n/a")
+            accuracy_norm = fewshot_exp_results[task_name].get("acc_norm", "n/a")
+            ppl = fewshot_exp_results[task_name].get("ppl", "n/a")
+            df = pd.concat([pd.DataFrame([[task_name, ppl, accuracy, accuracy_norm, num]], columns=df.columns), df], ignore_index=True)
+
+    with open(outfile, "w") as f:
+        df.to_csv(f, sep='\t', index=None)
+
```
vchiley, line 94: \n

llm/src/evaluation/eval.py
```diff
@@ -0,0 +1,145 @@
...
+    for num in num_fewshots:
+        results[num] = evaluate(
+                lm=model,
+                task_dict=task_dict,
+                num_fewshot=num,
+                limit=PARTIAL_EVAL_SAMPLE_SIZE if partial_eval_mode else None
+            )
+
+    return results
+
```
vchiley, line 70: \n

llm/src/evaluation/eval.py
```diff
@@ -0,0 +1,145 @@
...
+    Returns:
+        lm (lm_eval.base.LM): An lm_eval LM model wrapped around our constructed model
+    '''
+    model_components = MODEL_LOADERS[model_type](**model_ctor_args)
+    model_components.update({"batch_size": 4, "device": DEVICE.type})
+    lm = lm_eval_models.get_model("composer_llm").create_from_arg_string(
+        "", model_components
+    )
+    return lm
+
```
vchiley, line 42: \n

llm/src/evaluation/eval.py
```diff
@@ -0,0 +1,145 @@
...
+from lm_eval import tasks as lm_eval_tasks
+from lm_eval.evaluator import evaluate
+from src.evaluation.model_loading import MODEL_LOADERS
+
+RESULTS_DIR = f"{os.path.dirname(__file__)}/eval_reports"
+PARTIAL_EVAL_SAMPLE_SIZE = 40
+
+DEVICE = torch.device('cuda') if torch.cuda.is_available() else torch.device('cpu')
+JsonResults = Dict[int, Dict[str, Any]]
+
```
vchiley, line 23: \n

vchiley (commented):
Found a few random things which could be adjusted.
A few of my comments are just formatting type stuff which will eventually be found by a linter if not fixed now.
note: `\n` means add new line.

I haven't run `llm/src/evaluation/*` and am working off the assumption that you have.

llm/yamls/hf_causal_lm/gpt2-small.yaml
abhi-mosaic: I think the way it works is that `hf_config` can either be a string referencing a pretrained config living on `huggingface.co` or a local path to a fully defined HF Config. I think it's worth adding an example of that to the repo so people know it can be done.

Also, maybe the name for this param ought to be `config_name_or_path`, to make it more clear?

llm/src/mosaic_gpt.py
```diff
@@ -27,13 +27,26 @@ def __init__(self, cfg: Mapping[str, Any], device: str = None):
             batch_first=True,
             device=device,
         )
+        
         self.register_buffer(
-            "mask", torch.tril(torch.ones(cfg.max_seq_len, cfg.max_seq_len)))
+            "mask", nn.Transformer.generate_square_subsequent_mask(cfg.max_seq_len))
```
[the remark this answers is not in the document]
abhi-mosaic: Unfortunately I think this is only in torch1.13+, will make a note of this though when we are ready to upgrade.

https://github.com/pytorch/pytorch/blob/664058fa83f1d8eede5d66418abff6e20bd76ca8/torch/nn/modules/transformer.py#L152

llm/yamls/hf_causal_lm/gpt2-small.yaml
vchiley: hf_config is required here
https://github.com/mosaicml/benchmarks/blob/4d1a5a2b6e4934166371fcb6b136c2cb197e6532/llm/src/hf_causal_lm.py#L45

but i guess we can change that

llm/src/mosaic_gpt.py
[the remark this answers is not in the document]
abhi-mosaic: Resolved this by implementing the causal attention mask directly, with `device=...` passed through. I think when `device='meta'` it acts as a no-op.

llm/yamls/mosaic_gpt/760m.yaml
abhi-mosaic: Hmm i was trying to play it safe and just put values that will guaranteed not OOM... so basically if`device_train_microbatch_size` succeeds in train mode then it should also succeed in eval mode.... I think I'm gonna leave it as is for now just because I don't have the bandwidth to profile the true ceiling eval_batch_sizes that work for each config

llm/yamls/hf_causal_lm/gpt2-small.yaml
vchiley: Since we're using huggingface transformer's AutoConfig.from_pretrained, maybe we should just use the same arg name they use [here](https://huggingface.co/docs/transformers/v4.24.0/en/model_doc/auto#transformers.AutoConfig.from_pretrained)
ie `AutoConfig.from_pretrained( pretrained_model_name_or_path, **kwargs )`

llm/yamls/mosaic_gpt/350m.yaml
vchiley: https://github.com/mosaicml/benchmarks/pull/26#discussion_r1029940651

llm/src/mosaic_gpt.py
abhi-mosaic: Edited `self.mha` to be `self.mhsa` for Multiheaded Self Attention.

llm/src/evaluation/eval.py
vchiley: `--debug` triggers the pdb trace here: https://github.com/mosaicml/benchmarks/pull/26#discussion_r1029873927
since that was removed, the flag is useless

vchiley (approved):
LGTM

PR #28, opened 2022-11-22
Vitaliy/compare hf mosaic

vchiley:
This test compares HF's GPT2 and the MosaicGPT model for compatibility.

Usage: From `benchmarks/llm` run `pytest tests/compare_hf_v_mosaic_gpt.py`

Note: The test converts a HF GPT2 statedict into a MosaicGPT model statedict. The inverse of this functionality is useful if people want to use this repo to train a MosaicGPT2 model then publish the weights to HF Hub.


llm/tests/compare_hf_v_mosaic_gpt.py
```diff
@@ -0,0 +1,140 @@
...
+    with open('yamls/mosaic_gpt/125m.yaml') as f:
+        cfg = om.load(f)
+
+    # modify cfg for HF GPT2 compatibility
+    cfg.max_seq_len = 1024  # GPT2 seq len (GPT3 uses 2048)
+    cfg.model.max_seq_len = cfg.max_seq_len
+    cfg.train_loader.max_seq_len = cfg.max_seq_len
+    cfg.eval_loader.max_seq_len = cfg.max_seq_len
+    cfg.model.device = device
+    cfg.precision = 'fp16'
```
abhi-mosaic, line 28: I think this should be `amp` rather than `fp16` (silly Composer convention), but I also don't think it will affect anything as you are running the forwards pass yourself.

llm/tests/compare_hf_v_mosaic_gpt.py
```diff
@@ -0,0 +1,140 @@
...
+
+    # get 125m config
+    with open('yamls/mosaic_gpt/125m.yaml') as f:
+        cfg = om.load(f)
+
+    # modify cfg for HF GPT2 compatibility
+    cfg.max_seq_len = 1024  # GPT2 seq len (GPT3 uses 2048)
+    cfg.model.max_seq_len = cfg.max_seq_len
+    cfg.train_loader.max_seq_len = cfg.max_seq_len
+    cfg.eval_loader.max_seq_len = cfg.max_seq_len
```
abhi-mosaic, line 26: Maybe a more general comment, I think we can extract `cfg.model` and only edit those config values, as the rest of these like `cfg.train_loader.max_seq_len ` won't really be used.

llm/tests/compare_hf_v_mosaic_gpt.py
```diff
@@ -0,0 +1,140 @@
...
+
+    # for b in model.model.transformer.blocks:    b.resid_mlp_dropout.p = 0.0
+    # for b in hf_model.model.transformer.h:      b.mlp.dropout.p = 0.0
+
+    # for b in model.model.transformer.blocks:    b.resid_attn_dropout.p = 0.0
+    # for b in hf_model.model.transformer.h:      b.attn.resid_dropout.p = 0.0
+
+    # # attn_dropout is imtegrated into the FlashMHA kernel and must therefore be set to 0
+    # for b in model.model.transformer.blocks:    b.attn_dropout.p = 0.0 # set in cfg
+    for b in hf_model.model.transformer.h:      b.attn.attn_dropout.p = 0.0
```
abhi-mosaic, line 76: never seen this notation before without the indents!

llm/tests/compare_hf_v_mosaic_gpt.py
```diff
@@ -0,0 +1,140 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
+import warnings
+
+import torch
+from omegaconf import OmegaConf as om
+from composer.utils import reproducibility
+
+from main import build_composer_model
```
abhi-mosaic, line 11: I think this path has changed with the latest changes on main, should be something like `from src.model_registry import COMPOSER_MODEL_REGISTRY`, I think.

llm/tests/compare_hf_v_mosaic_gpt.py
```diff
@@ -0,0 +1,140 @@
...
+
+    # # attn_dropout is imtegrated into the FlashMHA kernel and must therefore be set to 0
+    # for b in model.model.transformer.blocks:    b.attn_dropout.p = 0.0 # set in cfg
+    for b in hf_model.model.transformer.h:      b.attn.attn_dropout.p = 0.0
+
+    hf_model.train()
+    model.train()
+
+    # UTIL: can be used to verify that models are not the same at init
+    # with torch.autocast(device_type='cuda', dtype=torch.float16):
```
abhi-mosaic, line 82: This seems valuable to keep, and we can add an assert that the outputs are not equal?

llm/tests/compare_hf_v_mosaic_gpt.py
```diff
@@ -0,0 +1,140 @@
...
+    with torch.autocast(device_type=device, dtype=torch.float16):
+        torch.manual_seed(cfg.seed)
+        hf_model_fwd = hf_model(batch)
+        torch.manual_seed(cfg.seed)
+        model_fwd = model(batch)
+
+    print(hf_model_fwd.mean().item(), model_fwd.mean().item())
+    print(hf_model_fwd, model_fwd)
+
+    # given dropout seeded the same way, the mean of the outputs is extreamly similar
```
abhi-mosaic, line 138:
```suggestion
    # given dropout seeded the same way, the mean of the outputs is extremely similar
```

vchiley:
> Does the final torch.allclose get any closer if we turn off dropout?

Not in any meaningful way. The differences come down to numerical precision propagating and accumulating thru all the layers.
I set rng before the fwd of each model. The resulting dropped idx are the same (given the dropout layer is implemented the same). The one place where this logic fails is in the FlashAttn layer where dropout is implemented inside FlashAttn and therefore the implementation is diff from nn.Dropout.

I checked and removing dropout makes virtually no difference.

Note:
- not setting rng before each fwd (so that dropout has diff behavior) makes a pretty big diff (when dropout is enabled)
- alternatively setting dropout to have diff values makes a pretty big diff too.

> It might even make sense to parameterize this test case with dropout=[0, 0.1] or something like that.

sure why not.

vchiley:
re: https://github.com/mosaicml/benchmarks/pull/28#pullrequestreview-1192226736

I don't think it matters either way...

llm/tests/compare_hf_v_mosaic_gpt.py
vchiley: There exists a universe, where by chance, the outputs, at init, are the same.
I'll add a warnings.warn if that is the case ie
```
    if hf_model_fwd.mean().allclose(model_fwd.mean()):
        warn_msg = f'WARNING: model_fwd ({model_fwd}) and hf_model_fwd ({hf_model_fwd}) are very close at init.'
        raise warnings.warn(warn_msg)
```

llm/tests/compare_hf_v_mosaic_gpt.py
vchiley: deprecated

bmosaicml (approved):

PR #36, opened 2022-11-29
Add codeowners for each existing benchmark

dblalock:
3-line change. Not adding owners for BERT until that PR is in.


bandish-shah (approved):
👍

PR #37, opened 2022-11-29
Add precommit linting + get repo passing all checks

dblalock:
This adds the following pre-commit checks:
- yapf
- pyright
- pycln
- isort
- inserting our license header into every python file
- docformatter
- pydocstyle
- yamllint
- yamlfmt

This is mostly a copy-pasted `.pre-commit-config.yaml`, `pyproject.toml`, and `.yamllint.yaml`  from the streaming repo. The main differences are:

1. I restricted the Python line width to 80 rather than 99. Since we expect users to fork this repo and add to the source code directly, figured it's safer to accommodate people who prefer short lines.
    1. If you want 100-char lines but the existing code only uses 80, it doesn't hurt you. But if you want 80-char lines and the existing code uses 100, you have tons of hideous line wrapping that makes using our examples painful.
3. I added [yamlfmt](https://github.com/jumanjihouse/pre-commit-hook-yamlfmt) since our yamls were nowhere near compliant with yamllint and I didn't want to manually change them all. The one subtlety here is that it demands that list items not have the bullet indented relative to the parent (looks fine, just slightly different than streaming + composer yamls).

The rest of the code changes are just autoformatting + a few manual interventions to appease the linters.

Deferred to follow-up PRs:
- Testing (these hooks only lint/format)
- CI


dblalock:
Mostly looking for @bandish-shah to review this; repo just auto-requested everyone since you're now codeowners of your respective repos. That said, if you want to quickly look at the changes in your directory to make sure none of the semantics changed, that would be great--wouldn't be surprised if at least one bug got introduced.

pyproject.toml
```diff
@@ -0,0 +1,491 @@
+
+# iSort
+[tool.isort]
+multi_line_output = 0
+line_length = 80
+skip = [ "env", "wandb", "runs", "build", "node_modules" ]
+
+# Pyright
+[tool.pyright]
+include = [
```
dakinggg, line 10: Probably want to change the include/exclude paths here?

pyproject.toml
```diff
@@ -0,0 +1,491 @@
...
+reportImplicitStringConcatenation = "warning"  # TODO: make this an error
+reportInvalidStubStatement = "error"
+reportIncompleteStub = "error"
+reportUnsupportedDunderAll = "error"
+reportUnusedCoroutine = "error"
+
+# Pytest
+[tool.pytest.ini_options]
+# By default, do not run remote tests
+addopts = "--cov=streaming --cov-fail-under=50 --codeblocks --strict-markers -m 'not daily and not remote' -ra --tb=native"
```
dakinggg, line 54: adjust `cov` to not point to streaming probably?

dakinggg (commented):
Did a quick scan, noticed that some files seem to have a Composer copyright and a Benchmarks copyright. Was this intentional?

pyproject.toml
dblalock: ah; I have this fixed in a follow-up PR but not this one. Will backport so pytest is never in a broken state.

bandish-shah (approved):
LGTM, linting configs look good.  I didn't take a thorough look at the source code fixes since we would rely on the linting tools for that anyways.

PR #38, opened 2022-11-29
fix torch attn init

vchiley:
JiraIssue: https://mosaicml.atlassian.net/browse/RESEARCH-304

### DIFF
--- a/llm/src/mosaic_gpt.py
+++ b/llm/src/mosaic_gpt.py
@@ -7,6 +7,7 @@ Inspired by https://github.com/karpathy/minGPT/blob/master/mingpt/model.py
 """
 
 import math
+from functools import partial
 from typing import Any, Mapping
 
 import torch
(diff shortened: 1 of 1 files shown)


bmosaicml:
Two questions: 1. should we be ensuring that the mask buffer is initialized in the param init function and 2. can you double check that the behavior of the mask buffer is the same here as the way its implemented in llm-foundry? https://github.com/mosaicml/llm-foundry/blob/0134c536cfeb74be4f61a3ef59c5671ff66a6b27/llm/mosaic_gpt.py#L31-L32

bmosaicml:
I also wanted to follow up because I believe the Multi bug you are resolving is related to this Jira I made https://mosaicml.atlassian.net/browse/RESEARCH-301?atlOrigin=eyJpIjoiMTI3MmI1OWU0ZWEyNGVlMGI0ZmI5Mjc1NmUxOGQ1NzEiLCJwIjoiaiJ9

I found that the issue with torch attention of having a zeroed out causal LM mask doesn't arise until around step 500, at which point the perplexity suddenly drops to 0

> JiraIssue: https://mosaicml.atlassian.net/browse/RESEARCH-304

vchiley:
This pr addresses [this issue](https://mosaicml.atlassian.net/browse/RESEARCH-304) reported [here](https://mosaicml.slack.com/archives/C049H1E9LN8/p1669668549880069).
What you're describing seems to be a different issue which is already addressed as part of [pull/26](https://github.com/mosaicml/benchmarks/pull/26).

[pull/26 comment](https://github.com/mosaicml/benchmarks/pull/26#issue-1455704448) says: "Fix attention mask issue for TorchCausalAttention when MosaicGPT is built with device='meta'"
Issue with attn mask was noted [here](https://mosaicml-community.slack.com/archives/C02FPBRB85P/p1669156361550079), abhi's fix is noted in thread [here](https://mosaicml-community.slack.com/archives/C02FPBRB85P/p1669173240386469?thread_ts=1669156361.550079&cid=C02FPBRB85P)

bmosaicml (approved):

PR #39, opened 2022-12-02
Get precommit checks passing for BERT benchmark

dblalock:
They currently pass for everything else. This PR makes the 5kloc in the BERT directory also satisfy our linters / formatters.

### DIFF
--- a/bert/yamls/main/mcloud_run.yaml
+++ b/bert/yamls/main/mcloud_run.yaml
@@ -25,7 +25,8 @@ command: |
 # Changes to `parameters` will be reflected in `/mnt/config/parameters.yaml`, which
 # is the config that `main.py` uses in the above command.
 parameters:
-  data_remote: &data_remote # Supply a path to your remote dataset
+  # Supply a path to your remote dataset and, optionally, local dataset
+  data_remote:
   data_local: &data_local /tmp/mds-cache/
 
   max_seq_len: &max_seq_len 128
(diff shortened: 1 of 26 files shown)


dakinggg (approved):
Didn't spot any logic issues in my pass, LGTM

bert/yamls/main/mcloud_run.yaml
```diff
@@ -45,7 +45,7 @@ parameters:
   train_loader:
     name: c4
     dataset:
-      remote: *data_remote
+      remote:                    # Supply a path to your remote dataset
```
alextrott16, line 48: Seems like the `*data_remote` got nuked here, which is probably not an intended consequence of replacing `null` with empty.

alextrott16 (changes requested):
LGTM but see comment about some of the YAML field-linkage formatting getting destroyed. I think those links should be added back.

bert/yamls/main/mcloud_run.yaml
dblalock: Good catch. Looks like a yamlfmt bug. Fixed via workaround.

alextrott16: Is it possible to have `parameters.data_remote: &data_remote` (so that the value is blank but the linker `&data_remote` is still present), so that we can keep `parameters.train_loader.dataset.remote: *data_remote` and `parameters.eval_loader.dataset.remote: *data_remote`?

If not, the `parameters.data_remote` makes it more tedious to fill in the YAML (whereas it was included to do the opposite).

alextrott16 (approved):
See suggestion in comment, but LGTM

PR #40, opened 2022-12-02
Update flash attn version

mvpatel2000:
Now that flash attn has tagged releases, we should move off frozen fork. We're doing the same in Composer image

### DIFF
--- a/llm/requirements.txt
+++ b/llm/requirements.txt
@@ -3,7 +3,7 @@ torchtext==0.13.1
 torch==1.12.1
 mosaicml==0.11.1
 mosaicml-streaming==0.1.2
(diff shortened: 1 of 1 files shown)


llm/requirements.txt
```diff
@@ -3,7 +3,7 @@ torchtext==0.13.1
 torch==1.12.1
 mosaicml==0.11.1
 mosaicml-streaming==0.1.2
-flash-attn @ git+https://github.com/mvpatel2000/flash-attention@main
+flash-attn @ git+https://github.com/HazyResearch/flash-attention.git@v0.2.2
```
abhi-mosaic, line 6:
```suggestion
flash-attn==0.2.2
```

abhi-mosaic (approved):
Made one suggestion. Otherwise LGTM.

PR #41, opened 2022-12-06
Raise error if global batch size not divisible by world size

bcui19:
Adding an error if global batch size is not divisible by world size, otherwise this will silently error. 

Addresses:
[CO-1463](https://mosaicml.atlassian.net/jira/software/c/projects/CO/issues/CO-1463?filter=myopenissues)

### DIFF
--- a/llm/main.py
+++ b/llm/main.py
@@ -62,7 +62,6 @@ def calculate_batch_size_info(global_batch_size, device_microbatch_size):
                          'as a result, the batch size will be truncated, please adjust `global_batch_size` '
                          f'to be divisible by world size, {dist.get_world_size()}.')
     device_batch_size = global_batch_size // dist.get_world_size()
-    print ("calculated batch size now is: ", device_batch_size * dist.get_world_size())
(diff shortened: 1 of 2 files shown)


llm/main.py
```diff
@@ -57,6 +57,10 @@ def build_scheduler(cfg):
 
 
 def calculate_batch_size_info(global_batch_size, device_microbatch_size):
+    if global_batch_size % dist.get_world_size() != 0:
+        raise ValueError(f'Global batch size {global_batch_size} is not divisible by {dist.get_world_size()} '
+                         'as a result, the batch size will be truncated, please adjust `global_batch_size` '
```
dakinggg, line 62: "will" -> "would"

dakinggg (commented):
Can you also add this error to bert benchmark too? https://github.com/mosaicml/benchmarks/blob/8a0e2d10d0a0fd9e54707503ca1c57275b4722e8/bert/main.py#L117

dakinggg (approved):
LGTM!

PR #47, opened 2022-12-16
Add makefile with command to run tests for each benchmark

dblalock:
Lets us run all the tests for each subdirectory in one command, while respecting each subdirectory's stated dependencies. Also makes all the tests pass, at least on a machine with a CUDA GPU.

Changes:
- Adds a Makefile similar to that in the science + streaming repos. Lets you `make style` to run precommit checks and `make test` to run all the tests.
- Applies precommit formatting to llm/tests/test_model.py
- Makes llm/tests/*.py files start with 'test_' to adhere to convention + make `pytest tests` work.
- Explicitly shuts down dataloader workers in {cifar,resnet}/tests/test_trainer.py. This makes these tests hang at the end for only like 2 minutes instead of 4 minutes.
- Fixes a bug / edge case in `llm/tests/test_c4_data_prep_script.py` that makes these tests never succeed again after you interrupt them once. (Just makes sure a dir gets deleted before they start instead of crashing)


Makefile
```diff
@@ -0,0 +1,44 @@
...
+# ================================================================
+# Main targets
+# ================================================================
+
+.PHONY: style
+style:  ## Apply autoformating and run style checks via pre-commit
+	pre-commit run --all-files
+
+# we don't test the BERT examples since there are no tests yet...
+.PHONY: style
```
dakinggg, line 35:
```suggestion
.PHONY: test
```

dakinggg (approved):
LGTM

PR #50, opened 2022-12-17
Dataloader

vchiley:
The current implementation of the c4 dataset can be used as a generic dataloader for text data. This PR modifies it to reflect that.
To sample from c4's val data:
`python src/text_data.py s3://my-bucket/my-copy-c4`
To sample from the-piles's val data:
`python src/text_data.py s3://my-bucket/my-copy-the-pile`


abhi-mosaic (commented):
Added a few comments, I think if we're making this generic now we may not need some of the functions?

vchiley:
deprecated in favor of #51

abhi-mosaic (approved):
LGTM!

PR #53, opened 2022-12-20
adding throughtput numbers and tools for getting numbers

vchiley:
Adding Mosaic_GPT's throughput / MFU numbers.
Includes scripts using mcli's python api for launching a bunch of runs / configs


llm/throughput/README.md
```diff
@@ -0,0 +1,165 @@
+# MosaicGPT Throughput, MFU, and HFU
+Throughput, MFU, and HFU for different sized MosaicGPT models trained on MCloud.
```
abhi-mosaic, line 2:
```suggestion
Throughput, MFU, and HFU for different sized MosaicGPT models trained on [MosaicML Cloud](https://www.mosaicml.com/cloud).
```

llm/throughput/README.md
```diff
@@ -0,0 +1,165 @@
...
+MFU and HFU are defined in https://arxiv.org/abs/2205.05198
```
abhi-mosaic, line 3: Before jumping into all the MFU stuff, can we have a 'Table of Contents' or intro paragraph explaining what the different scripts here are used for? `run_all_config.py` and `parse_logs.py` only make sense in the context of MosaicML Cloud so it would be good to address that

llm/throughput/README.md
```diff
@@ -0,0 +1,165 @@
...
+HFU is approximated using the fact that the networks use checkpointing for activation recomputation. This effectively requires an extra forward pass through the network.
+```
+hfu* = 4 * flops_per_seq * throughput / (gpu_num * GPU_AVAILABLE_FLOPS)
+hfu = (4 * flops_per_seq + 4 * attn_flops_per_seq) * throughput / (gpu_num * GPU_AVAILABLE_FLOPS)
+```
+
+These are approximations. Real HFU would be higher since it would include the FLOP usage of norm, act, residual, as well as account for ALL recomputation (the model uses FlashAttn which already does a recomputation in the fwd pass; in reality, the hfu attn multiplier should be 5 instaed of 3).
+
+Below we highlight some of the configurations, but the full tables can be found in [FULL_TABLES](./FULL_TABLES.md)
+
```
abhi-mosaic, line 27: Can we add one more sentence to the README.md, something about how "the configs here are not exhaustive, if there is a setting you are interested in, please reach out to [...] or if you already have access to MosaicML Cloud, try profiling it yourself like this: [EXAMPLE COMMAND]"

llm/throughput/README.md
```diff
@@ -0,0 +1,165 @@
...
+## A100 80GB
```
abhi-mosaic, line 28: The "highlights" are still quite long, do you think we should try to shorten them? Maybe like keeping the seqlen=[2048...16384] sweep but cutting down on the GPUType=[..] sweep? I'm not super opinionated here but it does feel like a lot even when looking on my laptop.

llm/throughput/run_all_configs.py
```diff
@@ -0,0 +1,306 @@
...
+        parameters['loggers'] =  {'wandb': {}}
+
+    return parameters
+
+
+def get_integrations(project, wandb=True):
+    integrations = [{
+        'integration_type': 'git_repo',
+        'git_repo': 'vchiley/benchmarks',
+        'git_branch': 'llm-thru-bauto',
```
abhi-mosaic, line 175: Ideally this should work?
```suggestion
        'git_repo': 'mosaicml/examples',
        'git_branch': 'main',
```

llm/throughput/run_all_configs.py
```diff
@@ -0,0 +1,306 @@
...
+import requests
+import yaml
+
+from mcli.sdk import RunConfig, create_run
+
+
+CLUSTER_INFO = {
+    # Cluster: [(gpu_type, max_gpus_per_run)],
+    'r1z1': [('a100_80gb',   8),],
+    'r7z2': [('a100_40gb', 128),],
```
abhi-mosaic, line 11: We should probably remove these, or put dummy names.

abhi-mosaic (approved):
Looks sweet! Approving for now as I think it would be good to get this merged quickly and update on the go. After torch 1.13 / composer 0.12 / Triton kernels we'll have to come back to this anyways..

abhi-mosaic:
I also have a feeling that we may be underselling the 1B and 3B models on 80GB cards, because they should work without activation checkpointing... but upgrading those #s can be a followup PR and I'll make a JIRA ticket to track.

llm/throughput/README.md
vchiley: cut down on a few

PR #56, opened 2022-12-21
Add MFU Monitoring to LLM

vchiley:
Adds MFU to SpeedMonitor
A generalized variant of this feature should eventually be integrated into Composer's SpeedMonitor callback, but this will work well for MosaicGPT in examples/llm.

Screenshot 2022-12-21 at 1 54 21 PM
Screenshot 2022-12-21 at 1 55 49 PM


llm/main.py
```diff
@@ -24,13 +26,30 @@ def build_logger(name, kwargs):
...
+        # the number of paramters is approximately the number of multiply-accumulates (MAC) in the network
+        # each MAC has 2 FLOPs - we multiply by 2 ie 2 * n_param
+        # this gets us FLOPs / token
+        params_flops_per_token = 2 * cfg.n_params
+        params_flops_per_seq = params_flops_per_token * cfg.max_seq_len
+        # there are 2 FLOPS per mac; there is A=Q*K^T and out=A*V ops (ie mult by 2)
+        attn_flops_per_seq = cfg.model.n_layers * 2 * 2 * (cfg.model.d_model * (cfg.max_seq_len ** 2))
+        return params_flops_per_seq + attn_flops_per_seq
+    
+    return None
```
abhi-mosaic, line 40: Can we co-locate this with the model definition? My gut feeling is that it makes more sense there than in the training script. Like what if it lived on `ComposerModel` and then maybe the `SpeedMonitor` searches for it and calls if it exists?

abhi-mosaic: This function could also maybe be defaulted in the `ComposerModel` base class! Like `6 *  sum(p.numel() for p in model.parameters())`.

llm/src/speed_monitor_w_mfu.py
```diff
@@ -0,0 +1,100 @@
...
+        batch_wct = state.timestamp.total_wct.total_seconds() - self.batch_start_wct
+
+        # Add the new element
+        self.batch_wct_buffer.append(batch_wct)
+        self.batch_num_samples_buffer.append(batch_num_samples)
+
+        # Log the throughput
+        if len(self.batch_num_samples_buffer) == self.window_size:
+            throughput = sum(self.batch_num_samples_buffer) / sum(self.batch_wct_buffer)
+            logger.log_metrics({'throughput/samples_per_sec': throughput})
```
abhi-mosaic, line 89: While we're here... do you want to add these? ;) 

```
throughput/batches_per_sec
throughput/samples_per_sec
throughput/tokens_per_sec
throughput/flops
throughput/device/batches_per_sec
throughput/device/samples_per_sec
throughput/device/tokens_per_sec
throughput/device/flops
throughput/device/mfu
```
(not attached to the naming, just wrote down what first comes to mind. Also would `mfu` be better categorized as a device-level metric?)

llm/main.py
vchiley: `6 *  sum(p.numel() for p in model.parameters())`
only works when you have MLP heavy network (ie wont work for CV networks).
This is also FLOPs per token not sample.

llm/src/speed_monitor_w_mfu.py
vchiley: to get `tokens_per_sec` we end up needing to get pretty specific ie `state.model.model.cfg.max_seq_len`.
This might not generalize well to other settings ie cv or other language models (non-mosaic specific). Is there a better place to get that number?

vchiley: Screenshot 2022-12-23 at 4 12 26 PM
Screenshot 2022-12-23 at 4 13 08 PM

Like this?

abhi-mosaic: > to get `tokens_per_sec` we end up needing to get pretty specific ie `state.model.model.cfg.max_seq_len`. This might not generalize well to other settings ie cv or other language models (non-mosaic specific). Is there a better place to get that number?

Yeah so there is technically a data spec function called `get_num_tokens_from_batch`, and so we should probably try to call this somehow.. but it doesn't have a smart default so we would need to add a custom data_spec inside `text_data.py`. I can dig into this a bit to see how much work it takes.

abhi-mosaic: So I think this line here: https://github.com/mosaicml/composer/blob/2ff5300171b3bd542c0422164ecaa79850869059/composer/callbacks/speed_monitor.py#L104

needs a duplicate line that calls `state.timestamp.token`. And then it's the responsibility of the dataloader + data_spec to properly report the `get_num_tokens_from_batch` to the Trainer. And that would remove the need to use the model's `max_seq_len`.

vchiley: yeah without a dataspec, state.timestamp.token outputs 0:
```
>>> state.timestamp.token
Time(0, TimeUnit.TOKEN)
```
but I guess `state.dataloader.dataset.max_seq_len` is a good place to find seq len (not sure why i didn't just look for it there earlier); we can still check if `max_seq_len` is part of the `state.dataloader.dataset` before adding token info to the logger.
(I guess this wouldn't work for ragged tensors which have variable seq len)
lmk if you still want to updt the dataloader to use a dataspec with the appropriate custom `get_num_tokens_in_batch` fn; its probably the only way to get real counts for variable seq lens if using ragged tensors.

llm/src/speed_monitor_w_mfu.py
```diff
@@ -0,0 +1,64 @@
...
+from collections import deque
+from typing import Any, Deque, Dict
+
+from composer.core import Callback, State
+from composer.loggers import Logger
+from composer.utils import dist
+
+from composer.callbacks import SpeedMonitor
+
+GPU_AVAILABLE_FLOPS = 312_000_000_000_000
```
abhi-mosaic, line 16: Hmmm I think we should probably have this be a kwarg to `SpeedMonitorMFU`, or better yet, have it be a type like `Union[float, str]` and you can type in `h100`, `a100`, `v100`, `t4`, etc. I think we're a bit lucky to have A100s but some users of this repo may be on alternative hardware!

llm/src/speed_monitor_w_mfu.py
```diff
@@ -0,0 +1,64 @@
...
+            if hasattr(state.dataloader.dataset, 'max_seq_len'):
+                # only applicable to seq data / models
+                logger.log_metrics({'throughput/tokens_per_sec': throughput * state.dataloader.dataset.max_seq_len})
+                logger.log_metrics({'throughput/device/tokens_per_sec': dev_throughput * state.dataloader.dataset.max_seq_len})
+
+            if hasattr(state.model, 'num_fwd_flops'):
+                flops = (3 * state.model.num_fwd_flops) * throughput
+                logger.log_metrics({'throughput/flops': flops})
+                device_flops = flops / world_size
+                logger.log_metrics({'throughput/device/flops': flops / world_size})
```
abhi-mosaic, line 54:
```suggestion
                logger.log_metrics({'throughput/device/flops': device_flops})
```

abhi-mosaic: Or make it `dev_flops` to match the `dev_` prefix above?

abhi-mosaic (approved):
Approving to unblock, but I think it's important that we don't just hardcode the `312e12` number, otherwise when people run this repo on alternative hardware they will get incorrect MFU values in their reporting.

llm/src/speed_monitor_w_mfu.py
abhi-mosaic: Future: we probably want Composer to autodetect and expose the device info.. seems pretty useful. WandB figured how to do it:
Screen Shot 2022-12-29 at 9 36 39 PM

vchiley: `torch.cuda.get_device_name()` outputs something like: `'NVIDIA A100-SXM4-40GB'`
using this and hardcoding these gpu type flop counts:
```
GPU_AVAILABLE_FLOPS = {
    # source: https://resources.nvidia.com/en-us-tensor-core/nvidia-tensor-core-gpu-datasheet
    'h100-sxm':     1.979e15 / 2,   # nvidia publishes spec sheet with a 2x sparsity factor
    'h100-pcie':    1.513e15 / 2,   # nvidia publishes spec sheet with a 2x sparsity factor
    # source: https://www.nvidia.com/content/dam/en-zz/Solutions/Data-Center/a100/pdf/nvidia-a100-datasheet-us-nvidia-1758950-r4-web.pdf
    'a100':         312e12,         # sxm and pcie have same flop counts
    # source: https://images.nvidia.com/content/technologies/volta/pdf/tesla-volta-v100-datasheet-letter-fnl-web.pdf
    'v100-sxm':     125e12,
    'v100-pcie':    112e12,
    # source: https://www.nvidia.com/content/dam/en-zz/Solutions/Data-Center/tesla-t4/t4-tensor-core-datasheet-951643.pdf
    't4':           65e12,          # sxm and pcie have same flop counts
}
```
Future: create something like a composer.gpu.constants tool for going thins

abhi-mosaic: Sorry to be pedantic, but I think we'll also need to hardcode these values per-precision type, and then look at the `state.precision` to figure out which value to use.

vchiley: updted to
```
GPU_AVAILABLE_FLOPS = {
    # source: https://resources.nvidia.com/en-us-tensor-core/nvidia-tensor-core-gpu-datasheet
    # nvidia publishes spec sheet with a 2x sparsity factor
    'h100-sxm':     {'fp64': 67e12, 'fp32': 67e12, 'tf32': 989e12 / 2, 'fp16': 1.979e15 / 2, 'amp_fp16': 1.979e15 / 2, 'bf16': 1.979e15 / 2, 'amp_bf16': 1.979e15 / 2, 'fp8': 3_958 / 2, 'int8': 3_958 / 2},
    'h100-pcie':    {'fp64': 51e12, 'fp32': 51e12, 'tf32': 756e12 / 2, 'fp16': 1.513e15 / 2, 'amp_fp16': 1.513e15 / 2, 'bf16': 1.513e15 / 2, 'amp_bf16': 1.513e15 / 2, 'fp8': 3_026 / 2, 'int8': 3_026 / 2},
    # source: https://www.nvidia.com/content/dam/en-zz/Solutions/Data-Center/a100/pdf/nvidia-a100-datasheet-us-nvidia-1758950-r4-web.pdf
    # sxm and pcie have same flop counts
    'a100':         {'fp64': 19.5e12, 'fp32': 19.5e12, 'tf32': 156e12, 'fp16': 312e12, 'amp_fp16': 312e12, 'bf16': 312e12, 'amp_bf16': 312e12},
    # source: https://images.nvidia.com/content/technologies/volta/pdf/volta-v100-datasheet-update-us-1165301-r5.pdf
    'v100-sxm':     {'fp64': 7.8e12, 'fp32': 15.7e12, 'fp16': 125e12, 'amp_fp16': 125e12},
    'v100-pcie':    {'fp64': 7e12, 'fp32': 14e12, 'fp16': 112e12, 'amp_fp16': 112e12},
    'v100s-pcie':   {'fp64': 8.2e12, 'fp32': 16.4e12, 'fp16': 130e12, 'amp_fp16': 130e12},
    # source: https://www.nvidia.com/content/dam/en-zz/Solutions/Data-Center/tesla-t4/t4-tensor-core-datasheet-951643.pdf
    # sxm and pcie have same flop counts
    't4':           {'fp32': 8.1e12, 'fp16': 65e12, 'amp_fp16': 65e12, 'int8': 130e12, 'int4': 260e12},
}
```

its a bit verbose but works. lmk if you had something else in mind.

abhi-mosaic (approved):
LGTM! Excited to have all this info auto generated.

PR #57, opened 2022-12-22
keep zip files when testing

samhavens:
fixes: calling `text_data.py` to test a dataset will delete zip files from your local dataset. if you then upload the data to s3 after testing it, your index will point to nonexistent zip files, causing errors when you try to use the data from s3.

### DIFF
--- a/llm/src/text_data.py
+++ b/llm/src/text_data.py
@@ -32,6 +32,8 @@ class StreamingTextDataset(Dataset):
         max_seq_len (int): The max sequence length of each sample.
         group_method (str): How to group text samples into token samples.
             Supports 'truncate' or 'concat'.
+        keep_zip (bool): Whether to keep or delete the compressed file when
+            decompressing downloaded shards.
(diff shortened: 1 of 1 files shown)


vchiley (approved):

llm/src/text_data.py
```diff
@@ -161,6 +164,7 @@ def build_text_dataloader(cfg: DictConfig, device_batch_size: int):
         tokenizer_name=cfg.dataset.tokenizer_name,
         max_seq_len=cfg.dataset.max_seq_len,
         group_method=cfg.dataset.group_method,
+        keep_zip=cfg.dataset.get('keep_zip', False),
```
abhi-mosaic, line 167: This makes me squint a bit as we're adding a default for this one field but not the others... but it also doesn't seem worth upgrading all the YAMLs just for this. There must be some kind of dataclass - based defaulting thing we could do in the future, but not worth it rn..

abhi-mosaic (approved):
LGTM!

PR #58, opened 2022-12-23
Triton&Flashibi in examples/llm

vchiley:
This pr integrates the triton implementation of FlashAttn (with ALiBi) into examples/llm.
Using this variant, memory usage is the same
Screenshot 2022-12-22 at 4 54 33 PM
it runs slightly faster
Screenshot 2022-12-22 at 4 51 15 PM
and with Alibi convergence is faster (and we converge to a lower final loss)
Screenshot 2022-12-22 at 4 51 08 PM
Screenshot 2022-12-22 at 4 51 23 PM

Meta-comment: MosaicGPT is currently structured so that the Causal Attn Bias is a member of the layer. It should really be a member of the model and should then be passed to all layers in their fwd pass. FlashAttn does not take a Bias / Mask so it does not suffer the memory penalty of having multiple masks. The Torch and Triton implementations do therefore are affected by the issue. @abhi-mosaic maybe we should create an issue to track this. We could make it part of this PR, but its slightly outside the scope.

Note: I tried a symlink `bert/src/flash_attn_triton.py -> llm/src/flash_attn_triton.py` but the result was a copy pasta; I'm not sure git supports symlinks...

### DIFF
--- a/llm/src/flash_attention.py
+++ b/llm/src/flash_attention.py
@@ -1,0 +1,121 @@
+# Copyright 2022 MosaicML Examples authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Triton implementation of Flash Attention Layer
+
+Modified from:
+https://raw.githubusercontent.com/HazyResearch/flash-attention/main/flash_attn/flash_attention.py
+"""
+
+import math
+import torch
+import torch.nn as nn
+from torch import Tensor
+from einops import rearrange
+from typing import List, Optional, Tuple, Union
+
+try:
+    from src.flash_attn_triton import flash_attn_qkvpacked_func, flash_attn_func
+except ImportError as e:
+    raise e
+
+
+class FlashAttention(nn.Module):
+    """Implement the scaled dot product attention with softmax.
+    Arguments
+    ---------
+        softmax_scale: The temperature to use for the softmax attention.
+                      (default: 1/sqrt(d_keys) where d_keys is computed at
+                      runtime)
+        attention_dropout: The dropout rate to apply to the attention
+                           (default: 0.0)
+    """
+    def __init__(self, num_heads, softmax_scale=None, attention_dropout=0.0, device=None, dtype=None):
+        super().__init__()
+        self.num_heads = num_heads
+        self.softmax_scale = softmax_scale
+        assert attention_dropout == 0
+        self.dropout_p = attention_dropout
+
+    def forward(
+        self,
+        qkv,
+        key_padding_mask: Optional[Tensor] = None,
+        attn_mask: Optional[Tensor] = None,
+        is_causal : bool = False,
+        need_weights: bool = False,
+        average_attn_weights: bool = True) -> Tuple[Tensor, Optional[Tensor]]:
+        """Implements the multihead softmax attention.
+        Arguments
+        ---------
+            qkv: The tensor containing the query, key, and value. (B, S, 3, H, D) if key_padding_mask is None
+                if unpadded: (nnz, 3, h, d)
+            key_padding_mask: If specified, a mask of shape :math:`(N, S)` indicating which elements within ``key``
+                to ignore for the purpose of attention (i.e. treat as "padding"). For unbatched `query`, shape should be :math:`(S)`.
+                Binary and byte masks are supported.
+                For a binary mask, a ``True`` value indicates that the corresponding ``key`` value will be ignored for
+                the purpose of attention. For a float mask, it will be directly added to the corresponding ``key`` value.
+            attn_mask: If specified, a 2D or 3D mask preventing attention to certain positions. Must be of shape
+                :math:`(L, S)` or :math:`(N\cdot\text{num\_heads}, L, S)`, where :math:`N` is the batch size,
+                :math:`L` is the target sequence length, and :math:`S` is the source sequence length. A 2D mask will be
+                broadcasted across the batch while a 3D mask allows for a different mask for each entry in the batch.
+                Binary, byte, and float masks are supported. For a binary mask, a ``True`` value indicates that the
+                corresponding position is not allowed to attend. For a byte mask, a non-zero value indicates that the
+                corresponding position is not allowed to attend. For a float mask, the mask values will be added to
+                the attention weight.
+            is_causal: If specified, applies a causal mask as attention mask. Mutually exclusive with providing attn_mask.
+                Default: ``False``.
+            need_weights: If specified, returns ``attn_output_weights`` in addition to ``attn_outputs``.
+                Default: ``True``.
+            average_attn_weights: If true, indicates that the returned ``attn_weights`` should be averaged across
+                heads. Otherwise, ``attn_weights`` are provided separately per head. Note that this flag only has an
+                effect when ``need_weights=True``. Default: ``True`` (i.e. average weights across heads)
+        """
+        assert not need_weights and not average_attn_weights
+        assert qkv.dtype in [torch.float16, torch.bfloat16]
+        assert qkv.is_cuda
+
+        if key_padding_mask is not None and key_padding_mask.bool().logical_not().any():
+            raise NotImplementedError(f'assumes key_padding_mask is taken care of by attn_mask')
+        qkv = rearrange(qkv,
+                        'b s (t h d) -> b s t h d',
+                        t=3,
+                        h=self.num_heads)
+
+        attn_output = flash_attn_qkvpacked_func(qkv, attn_mask, is_causal, None)
+        output = rearrange(attn_output, 'b s h d -> b s (h d)')
+        return output, None
+
+
+class FlashMHA(nn.Module):
+
+    def __init__(self, embed_dim, num_heads, bias=True, batch_first=True, attention_dropout=0.0,
+                 causal=False, device=None, dtype=None, **kwargs) -> None:
+        assert batch_first
+        factory_kwargs = {'device': device, 'dtype': dtype}
+        super().__init__()
+        self.embed_dim = embed_dim
+        self.causal = causal
+
+        self.num_heads = num_heads
+        assert self.embed_dim % num_heads == 0, "self.kdim must be divisible by num_heads"
+        self.head_dim = self.embed_dim // num_heads
+        assert self.head_dim % 8 == 0 and self.head_dim <= 128, "Only support head_dim <= 128 and divisible by 8"
+
+        self.Wqkv = nn.Linear(embed_dim, 3 * embed_dim, bias=bias, **factory_kwargs)
+        self.inner_attn = FlashAttention(num_heads=num_heads, attention_dropout=attention_dropout, **factory_kwargs)
+        self.out_proj = nn.Linear(embed_dim, embed_dim, bias=bias, **factory_kwargs)
+
+    def forward(self, x, key_padding_mask=None, attn_mask=None, need_weights=False):
+        """x: (batch, seqlen, hidden_dim) (where hidden_dim = num heads * head dim)
+        key_padding_mask: bool tensor of shape (batch, seqlen)
+        """
+        qkv = self.Wqkv(x)
+        context, attn_weights = self.inner_attn(
+            qkv,
+            key_padding_mask=key_padding_mask,
+            attn_mask=attn_mask,
+            is_causal=self.causal,
+            need_weights=need_weights,
+            average_attn_weights=False)
+        return self.out_proj(context), attn_weights
--- a/llm/src/flash_attention.py
+++ b/llm/src/flash_attention.py
@@ -82,7 +82,7 @@ class FlashAttention(nn.Module):
                         t=3,
                         h=self.num_heads)
 
-        attn_output = flash_attn_qkvpacked_func(qkv, attn_mask, is_causal, None)
+        attn_output = flash_attn_qkvpacked_func(qkv, attn_mask, is_causal, self.softmax_scale)
         output = rearrange(attn_output, 'b s h d -> b s (h d)')
         return output, None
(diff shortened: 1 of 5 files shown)


llm/requirements.txt
```diff
@@ -4,6 +4,8 @@ torch==1.12.1
 mosaicml==0.11.1
 mosaicml-streaming==0.1.2
 flash-attn==0.2.2
+# need a newer version of triton
+triton==2.0.0.dev20221103
```
abhi-mosaic, line 8: any particular reason for this version vs. `2.0.0.dev20221202`

llm/src/mosaic_gpt.py
```diff
@@ -149,20 +231,18 @@ def __init__(self, cfg: DictConfig):
...
-                                 device=cfg.device),
-                wpe=nn.Embedding(cfg.max_seq_len,
-                                 cfg.d_model,
-                                 device=cfg.device),
-                emb_drop=nn.Dropout(cfg.emb_pdrop),
-                blocks=nn.ModuleList([
+        
+        self.alibi = cfg.get('alibi', False)
+        
+        self.transformer = nn.ModuleDict({"wte": nn.Embedding(cfg.vocab_size, cfg.d_model, device=cfg.device)})
```
abhi-mosaic, line 237:
```suggestion
        self.transformer = nn.ModuleDict({'wte': nn.Embedding(cfg.vocab_size, cfg.d_model, device=cfg.device)})
```

llm/src/mosaic_gpt.py
```diff
@@ -178,7 +258,9 @@ def forward(self,
                            device=input_ids.device).unsqueeze(0)
 
         tok_emb = self.transformer.wte(input_ids)  # type: ignore
-        pos_emb = self.transformer.wpe(pos)  # type: ignore
+        pos_emb = 0
```
abhi-mosaic, line 261: does the adding 0 add any overhead? Idk if PyTorch is smart enough to skip

abhi-mosaic (approved):
LGTM! 

Acknowledged on the `attn_bias` buffer, lets refactor that in a followup PR.

Also, we should add CI to verify that certain files are the same across `examples/[llm,bert]` but this can come in the new year. Also, we may end up refactoring into a common/ folder and avoid this issue.

llm/requirements.txt
vchiley: copy pasta [req from bert](https://github.com/mosaicml/examples/blob/main/bert/requirements.txt#L9)

vchiley:
waiting for https://github.com/mosaicml/composer/pull/1854

llm/requirements.txt
```diff
@@ -4,6 +4,8 @@ torch==1.13.0
 mosaicml==0.12.0
 mosaicml-streaming==0.1.2
 flash-attn==0.2.2
+# need a newer version of triton
+triton==2.0.0.dev20221103
```
abhi-mosaic, line 8:
```suggestion
triton==2.0.0.dev20221202
```

Any reason not to use the latest version?

llm/src/mosaic_gpt.py
```diff
@@ -113,11 +122,15 @@ class GPTBlock(nn.Module):
...
     def __init__(self, cfg: DictConfig, device: Optional[str] = None):
         super().__init__()
+        if cfg.get('alibi', False):
+            assert 'triton' in cfg.attn_impl, 'Only triton kernel supports alibi'
         self.ln_1 = nn.LayerNorm(cfg.d_model, device=device)
         if cfg.attn_impl == 'torch':
             self.causal_attn = TorchCausalAttention(cfg, device)
         elif cfg.attn_impl == 'flash':
             self.causal_attn = FlashCausalAttention(cfg, device)
+        elif 'triton' in cfg.attn_impl:
```
abhi-mosaic, line 132: Could you elaborate why we're looking for `'triton' in cfg.attn_impl` rather than a fixed key like `cfg.attn_impl == 'triton'`?

llm/src/mosaic_gpt.py
```diff
@@ -145,60 +159,167 @@ def __init__(self, cfg: DictConfig):
         super().__init__()
         assert cfg.name == 'mosaic_gpt', f'Tried to build MosaicGPT model with cfg.name={cfg.name}'
         self.cfg = cfg
+        self.alibi = cfg.get("alibi", None)
```
abhi-mosaic, line 162:
```suggestion
        self.alibi = cfg.get("alibi", False)
```

llm/src/mosaic_gpt.py
```diff
@@ -145,60 +159,167 @@ def __init__(self, cfg: DictConfig):
...
+        #   mask to your attentions. See https://pytorch.org/docs/stable/generated/torch.nn.MultiheadAttention.html#torch.nn.MultiheadAttention.forward
+        #   Basically True/-inf indicates tokens we do not want to attend to.
+        #
+        # 2. This is is the exact opposite behavior of Huggingface's tokenizers, which use the convention that True denotes tokens
+        #   we do want to attend to. See https://huggingface.co/docs/transformers/glossary#attention-mask
+        if self._attn_mask_initialized:
+            return self.attn_mask
+        
+        if not S:
+            S = x.size(1)
```
abhi-mosaic, line 263: Is `S` needed here?

llm/src/flash_attn_triton.py
```diff
@@ -0,0 +1,1112 @@
+# Copyright 2022 MosaicML Examples authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Triton implementation of Flash Attention.
```
abhi-mosaic, line 4: Just want to double check on this file, is it a direct copy of https://github.com/HazyResearch/flash-attention/blob/main/flash_attn/flash_attn_triton.py?

And if so, is it up to date (last change was 2 weeks ago)?

llm/src/mosaic_gpt.py
```diff
@@ -30,40 +30,16 @@ def __init__(self, cfg: DictConfig, device: Optional[str] = None):
...
-            self._fill_causal_attn_mask()
-            self.mask_initialized = True
-
-        return self.mhsa(x,
-                         x,
-                         x,
-                         attn_mask=self.mask,
-                         key_padding_mask=~key_padding_mask,
-                         need_weights=True)
+    def forward(self, x, key_padding_mask, attn_mask=None):
```
abhi-mosaic, line 35: Does this mean Torch-native attention now works with torch 1.13? I had this listed as a bug but it looks like you may have fixed it? 👀

abhi-mosaic (approved):
I made some minor comments, I think it would be good to have someone with more familiarity with ALIBI (@growlix , @alextrott16 )? to look over the attn_mask initialization. I wouldn't be able to tell if something is wrong.

The convergence improvement looks promising though! Maybe that's enough to support merging. Leaving it up to you @vchiley. 

Also, as we have added new features to `MosaicGPT`, the class has gotten quite big. I'll add a JIRA issue to refactor some of this, add comments, maybe move param_init_fn into a separate class, etc. It's getting harder to read and I want to make sure that when new users see the codebase for the first time, it's not too overwhelming.

llm/requirements.txt
vchiley: I copied it from the [bert requirements](https://github.com/mosaicml/examples/blob/main/bert/requirements.txt#L8)

llm/src/flash_attn_triton.py
vchiley: this is from the [bert dir](https://github.com/mosaicml/examples/blob/main/bert/src/flash_attn_triton.py)

I guess it points to the need for a unified code dir.

llm/src/mosaic_gpt.py
vchiley: just me being dumb, I'll update it to follow convention.

llm/src/mosaic_gpt.py
vchiley: nope

llm/src/mosaic_gpt.py
vchiley: not sure where the bug was / wasn't but [this test](https://github.com/vchiley/examples/blob/tritonFlashibiFaster/llm/tests/test_hf_v_mosaic_gpt.py#L18) checks the hf model against mosaic_gpt with torch attn and it passes.

Also the mask generation (including the comments you made about the mask generation) I moved to [where I generate the mask](https://github.com/vchiley/examples/blob/tritonFlashibiFaster/llm/src/mosaic_gpt.py#L250).

vchiley:
cc @abhi-mosaic 
> refactor

Probably a good idea if / when we create a 'common' folder which has (handles) attn, flash attn, triton flash attn, alibi etc

llm/src/flash_attn_triton.py
alextrott16: @abhi-mosaic Yes, but we have not been incorporating changes to that original implementation into the BERT implementation.
Agreed that a location for unified code would be advantageous. I support the choice to re-use the BERT code for now, since it will be easier to refactor things if it's just a matter of reducing redundancy.

alextrott16 (approved):
Overall: I think alibi is correctly implemented, so I'm approving on that basis.

I'll add, though, that the general implementation of the 3 attention versions makes this a bit hard to read. I trust that there are good reasons to structure the code this way, but I think it warrants some detailed comments around the code in `MosaicGPT` where the attention mask is built. Also, in the attention classes themselves, it would help to add comments to explain how the associated methods will be used.

Last note: perhaps due to being a bit hard to parse, I'm not totally clear on how much re-building of the attention mask happens on each forward call. I'm assuming that re-building is minimal (e.g., just handling batch-specific masks due to padding). If that's not the case, I can provide a more fine-grained review on possible ways to re-use pre-computed masks.

vchiley:
@alextrott16 
> how much re-building of the attention mask happens on each forward call

The generic and alibi attn masks are built and part of the model (`shape: 1 x 1 x 1 x SeqLen` or if alibi is used `shape: 1 x heads x 1 x SeqLen`; broadcasted to `shape: 1 x 1 x SeqLen x SeqLen` internally in the kernel; `causal=True` takes care of the causal portion of the mask)
These masks are only extended to `shape: BatchSize x 1 x 1 x SeqLen` when `key_padding_mask` says some of the tokens are invalid (only happens at eval since durring training `group_method: concat`).
As a result, yes "re-building is minimal", I "re-use pre-computed masks" as much as possible.

llm/src/flash_attention.py
```diff
@@ -0,0 +1,121 @@
...
+                      (default: 1/sqrt(d_keys) where d_keys is computed at
+                      runtime)
+        attention_dropout: The dropout rate to apply to the attention
+                           (default: 0.0)
+    """
+    def __init__(self, num_heads, softmax_scale=None, attention_dropout=0.0, device=None, dtype=None):
+        super().__init__()
+        self.num_heads = num_heads
+        self.softmax_scale = softmax_scale
+        assert attention_dropout == 0
```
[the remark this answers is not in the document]
vchiley: This entire file is a copy pasta from [hazy research](https://github.com/HazyResearch/flash-attention/blob/main/flash_attn/flash_attention.py); then modify it to use the triton kernel instead of the cuda flash attn kernel.
I kept it to keep the module signature the same, but to your point, I should probably just remove it and move the assertion higher in the stack.

llm/src/mosaic_gpt.py
```diff
@@ -86,10 +90,80 @@ def __init__(self, cfg: DictConfig, device: Optional[str] = None):
...
+    alibi_bias = torch.arange(1 - seq_len, 1, dtype=dtype, device=device).view(1, 1, 1, seq_len)
+    if full:
+        # generate 1 x Heads x SeqLen x SeqLen alibi bias mask
+        # otherwise the mask is 1 x Heads x 1 x SeqLen (which is braodcasted up to the approproate size)
+        alibi_bias = alibi_bias - torch.arange(1 - seq_len, 1, dtype=dtype, device=device).view(1, 1, seq_len, 1)
+        alibi_bias.abs_().mul_(-1)  # since we're using causal flag, this isn't really needed, but why not include it
+
+    m = torch.arange(1, n_heads + 1, dtype=dtype, device=device)
+    m.mul_(alibi_bias_max / n_heads)
+    alibi_bias = alibi_bias * (1. / (2 ** m.view(1, n_heads, 1, 1)))
```
growlix, line 165: I didn't spend enough time doing the mental computations to figure out how this obtains the same values for _m_ as in the Composer implementation/paper, but your results look good so I'm not too concerned.

growlix (approved):
Great work! Two minor observations. Otherwise I endorse Alex and Abhi's reviews.

llm/src/flash_attn_triton.py
dakinggg: perhaps we can link to the specific commit it is up to date with, so we don't lose track of that.

dakinggg: It would also be good to call out with comments anywhere that we have made changes. Not sure what the longer term plan is for keeping this in sync (or not) with hazy research.

llm/src/mosaic_gpt.py
vchiley: Last sentence of 2nd paragraph of section 3 of [the alibi paper](https://arxiv.org/abs/2108.12409) goes over the details. This is my interpretation and convergence works well.

llm/src/flash_attn_triton.py
vchiley: I symlinked to the BERT version of this file (but git doesn't respect symlinks)
changes can be see in this PRs changes to `bert/src/flash_attn_triton.py`

PR #60, opened 2023-01-03
Add optimization logger

bmosaicml:
This will allow us to use the optimization logger during training aid loss spike detection.

Monitoring the change with this run gpt-760m-weight-tying-LczKSX

### DIFF
--- a/llm/main.py
+++ b/llm/main.py
@@ -6,7 +6,7 @@ import sys
 import warnings
 
 from composer import Trainer, algorithms
-from composer.callbacks import LRMonitor, MemoryMonitor
+from composer.callbacks import LRMonitor, MemoryMonitor, OptimizerMonitor
 from composer.loggers import WandBLogger
 from composer.optim import DecoupledAdamW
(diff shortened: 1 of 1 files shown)


llm/main.py
```diff
@@ -34,6 +34,10 @@ def build_callback(name, kwargs):
         return SpeedMonitorMFU(
             window_size=kwargs.get('window_size', 1),
             gpu_flops_available=kwargs.get('gpu_flops_available', None))
+    elif name == 'optimizer_monitor':
+        return OptimizerMonitor(
+            log_optimizer_metrics=kwargs.get('log_optimizer_metrics'),
```
vchiley, line 39: Do we want to set a default ie `log_optimizer_metrics=kwargs.get('log_optimizer_metrics', DEFAULT)`?
(The [default in the callback definition](https://github.com/mosaicml/composer/blob/dev/composer/callbacks/optimizer_monitor.py#L85) is `True`)

vchiley (approved):
lgtm

PR #51, opened 2022-12-20
Streaming v0.2.1

vchiley:
~~updates examples/llm dataset / dataloader to use [Streaming v0.2.1](https://github.com/mosaicml/streaming/releases/tag/v0.2.1)~~
waiting for Streaming v0.2.2

PR #61, opened 2023-01-06
Upgrade `mosaicml-streaming`

abhi-mosaic:
This PR replaces #51.

The main changes here are:
* using `mosaicml-streaming==0.2.2`
* bugfix in SpeedMonitorMFU
* write out all of SpeedMonitorMFU rather than subclass, because otherwise it is hard for a new user to follow the logic
* trying to find a better balance of defaults vs. explicit hparams in YAMLs
* turning off progress bar in YAMLs, it's just too annoying with MCLOUD, users can re-enable if they want it

I've edited `125m.yaml` only with these changes, and after feedback, I will edit all the remaining YAMLs to match.


llm/src/speed_monitor_w_mfu.py
```diff
@@ -61,11 +61,7 @@ def get_gpu_flops_available(state):
         except:
             gpu_flops_available = None
 
-    if gpu_flops_available is not None:
-        print(
-            f'Using {gpu_flops_available=} when calculate MFU (assumes {state.precision.value} precision using {torch.cuda.get_device_name()}).'
-        )
-    else:
```
abhi-mosaic, line 68: @vchiley the reason I'm removing this for now is b/c it was printing out after the first batch of each run and it looks kinda messy for demos

llm/requirements.txt
```diff
@@ -2,7 +2,7 @@ torchvision==0.14.0
 torchtext==0.14.0
 torch==1.13.0
 mosaicml==0.12.0
-mosaicml-streaming==0.1.2
+mosaicml-streaming @ git+https://github.com/mosaicml/streaming.git@main
```
abhi-mosaic, line 5: To be upgraded

llm/src/text_data.py
```diff
@@ -11,69 +11,76 @@
...
+                 download_timeout: float = 120,
+                 validate_hash: Optional[str] = None,
+                 shuffle_seed: int = 9176,
+                 num_canonical_nodes: Optional[int] = None,
                  batch_size: Optional[int] = None):
+
         # Validation
-        if split not in ['train', 'val']:
-            raise ValueError(
-                f"split='{split}' must be one of ['train', 'val'].")
```
abhi-mosaic, line 61: @vchiley this should fix the issue you had with custom `split` names.

llm/main.py
```diff
@@ -5,17 +5,19 @@
...
 from composer.callbacks import LRMonitor, MemoryMonitor, OptimizerMonitor
 from composer.loggers import WandBLogger
 from composer.optim import DecoupledAdamW
 from composer.optim.scheduler import (ConstantWithWarmupScheduler,
                                       CosineAnnealingWithWarmupScheduler)
 from composer.utils import dist, reproducibility
-from omegaconf import OmegaConf as om
-from src.text_data import build_text_dataloader
-from src.model_registry import COMPOSER_MODEL_REGISTRY
-from src.speed_monitor_w_mfu import SpeedMonitorMFU
```
vchiley, line 18: This is a finicky (personal) style thing but I feel like it should generally be
```
import base python things ie sys, warning

import pip installed libraries (including composer stuff)

import everything implemented in this proj
```

u've moved the `import everything implemented in this proj` portion to before composer imports (where composer is a pip installed lib)

llm/main.py
```diff
@@ -204,21 +206,19 @@ def main(cfg):
         schedulers=scheduler,
         max_duration=cfg.max_duration,
         eval_interval=cfg.eval_interval,
-        eval_subset_num_batches=cfg.eval_loader.get('eval_subset_num_batches',
-                                                    -1),
+        eval_subset_num_batches=cfg.eval_subset_num_batches,
```
vchiley, line 209: Assuming every config using this code will set `eval_subset_num_batches` is a strong assumption.
Most ppl would expect it to be `-1` ie eval on the entire val set.

llm/main.py
```diff
@@ -204,21 +206,19 @@ def main(cfg):
...
-        algorithms=algos,
-        device_train_microbatch_size=cfg.get('device_train_microbatch_size', 'auto'),
+        algorithms=algorithms,
+        device_train_microbatch_size=cfg.device_train_microbatch_size,
         fsdp_config=fsdp_config,  # type: ignore
         save_folder=cfg.get('save_folder', None),
         save_interval=cfg.get('save_interval', '1000ba'),
-        save_num_checkpoints_to_keep=cfg.get('save_num_checkpoints_to_keep',
-                                             -1),
+        save_num_checkpoints_to_keep=cfg.get('save_num_checkpoints_to_keep', -1),
```
vchiley, line 221: I feel like this is one of those edits a linter made ie 
`        save_num_checkpoints_to_keep=cfg.get('save_num_checkpoints_to_keep', -1),`
is 81 chars long so a linter split the line. You're undoing it. When whoever (ie Davis) runs a linter in the future, they'll redo it.

feel free to ignore this cmt.

llm/src/speed_monitor_w_mfu.py
```diff
@@ -74,28 +69,57 @@ def get_gpu_flops_available(state):
...
     ):
-        super().__init__(window_size=window_size)
-        self.set_gpu_flops_available = False
+        # Track the batch num samples and wct to compute throughput over a window of batches
+        self.batch_start_num_samples = 0
+        self.batch_start_wct = 0.0
+        self.batch_wct_buffer: Deque[float] = deque(maxlen=window_size)
+        self.batch_num_samples_buffer: Deque[int] = deque(maxlen=window_size)
+        self.window_size = window_size
         self.gpu_flops_available = gpu_flops_available
```
vchiley, line 86: I know you're fixing my mistakes but should't this just be 
```
if gpu_flops_available:
    self.gpu_flops_available = gpu_flops_available
else:
    self.gpu_flops_available = get_gpu_flops_available(state)
```
instead of setting `self.gpu_flops_available` in `before_dataloader()`

llm/src/speed_monitor_w_mfu.py
```diff
@@ -74,28 +69,57 @@ def get_gpu_flops_available(state):
...
+        self.total_eval_wct = state['total_eval_wct']
+
+    def before_dataloader(self, state: State, logger: Logger) -> None:
+        del logger  # unused
+        self.batch_start_wct = state.timestamp.total_wct.total_seconds()
+        self.batch_start_num_samples = int(state.timestamp.sample)
+
+        # Get available GPU FLOPS
+        if not self.gpu_flops_available:
             self.gpu_flops_available = get_gpu_flops_available(state)
```
vchiley, line 120: see cmt on line 86

llm/src/text_data.py
```diff
@@ -11,69 +11,76 @@
...
-                 shuffle: bool,
-                 prefetch: int,
-                 tokenizer_name: str,
-                 max_seq_len: int,
+                 remote: Optional[str] = None,
+                 split: Optional[str] = None,
+                 shuffle: bool = False,
+                 predownload: int = 100_000,
+                 tokenizer_name: str = 'gpt2',
+                 max_seq_len: int = '2048',
```
vchiley, line 56: should the default be a string?

llm/src/text_data.py
```diff
@@ -11,69 +11,76 @@
...
                  group_method: str = 'truncate',
                  keep_zip: bool = False,
```
vchiley, line 58:
```suggestion
                 keep_zip: Optional[bool] = None,
```
keep is consistent with [streaming ds defaults](https://github.com/mosaicml/streaming/blob/main/streaming/base/dataset.py#L135)

llm/src/text_data.py
```diff
@@ -11,69 +11,76 @@
...
                  local: str,
-                 split: str,
-                 shuffle: bool,
-                 prefetch: int,
-                 tokenizer_name: str,
-                 max_seq_len: int,
+                 remote: Optional[str] = None,
+                 split: Optional[str] = None,
+                 shuffle: bool = False,
+                 predownload: int = 100_000,
```
vchiley, line 54:
```suggestion
                 predownload: Optional[int] = 100_000,
```
keep is consistent with [streaming ds defaults](https://github.com/mosaicml/streaming/blob/main/streaming/base/dataset.py#L134)

llm/src/text_data.py
```diff
@@ -156,15 +163,20 @@ def __len__(self) -> Optional[int]:
...
+        split=cfg.dataset.split,
         shuffle=cfg.dataset.shuffle,
-        prefetch=cfg.dataset.prefetch,
-        tokenizer_name=cfg.dataset.tokenizer_name,
-        max_seq_len=cfg.dataset.max_seq_len,
-        group_method=cfg.dataset.group_method,
+        predownload=cfg.dataset.get('predownload', 100_000),
+        tokenizer_name=cfg.dataset.get('tokenizer_name', 'gpt2'),
+        max_seq_len=cfg.dataset.get('max_seq_len', 2048),
+        group_method=cfg.dataset.get('group_method', 'truncate'),
```
vchiley, line 173: I feel like `max_seq_len` and `group_method ` shouldn't be set using a default...
we call this the Danger Zone

vchiley (approved):
Few random comments, but generally lgtm

llm/main.py
abhi-mosaic: I think i need to install the linter locally to fix this, will do that

llm/src/speed_monitor_w_mfu.py
abhi-mosaic: The trouble is `state` is not available on `__init__` :(

llm/src/text_data.py
abhi-mosaic: Good catch!

llm/src/text_data.py
abhi-mosaic: Ok, I will change to match Streaming defaults, but I'm definitely squinting a bit at the `Optional[bool]`...

llm/src/text_data.py
abhi-mosaic: Ok, let me see what it looks like to move a couple of these args up to be non-defaulted. I think it just makes the arg order a little less pretty, but it makes sense. I feel like `tokenizer_name` also falls in the same category.

llm/main.py
abhi-mosaic: I think this is related to the "where defaults go" issue. I think I'd rather explicitly put it into every YAML as -1, because that way as a new user you can see right away in both `main.py` and in the YAMLs that it's an option. I think we've had a couple customers not be able to find this flag before so I'm trying to make it more visible all throughout

abhi-mosaic: I guess I can revert it for now to not break existing YAMLs, and we can make a JIRA ticket to talk about defaults after?

llm/main.py
abhi-mosaic: Got it, I think this may be a linter/import sorting thing that I haven't configured correctly. Will try to fix it...

abhi-mosaic:
@vchiley could you take another look at `125m.yaml` and let me know if it looks good to you? If so, I will use that as a template to update all the other GPT and BERT YAMLs, do some final sanity checks, and merge.

vchiley:
yaml looks good

Q1: should we updt the dataloader to use [StreamingDataLoader](https://streaming.docs.mosaicml.com/en/stable/api_reference/generated/streaming.StreamingDataLoader.html) instead of the torch dataloader [here](https://github.com/mosaicml/examples/blob/abhi/streaming-023/common/text_data.py#L188)

Q2: [Dataset timeout defaults to 120](https://github.com/mosaicml/examples/blob/abhi/streaming-023/common/text_data.py#L176); y not use 60 ie the [StreamingDataset default](https://streaming.docs.mosaicml.com/en/stable/api_reference/generated/streaming.StreamingDataset.html)? We're leaving the [dataloader timeout](https://github.com/mosaicml/examples/blob/abhi/streaming-023/common/text_data.py#L197) at 0 even tho Dataset timeout is 120? (just making sure this was intended)

llm/yamls/mosaic_gpt/125m.yaml
```diff
@@ -33,37 +34,31 @@ tokenizer:
 # Dataloaders
 train_loader:
   dataset:
-    remote: *data_remote
     local: *data_local
+    remote: *data_remote
     split: train
     shuffle: true
-    prefetch: 1_000_000
+    predownload: 1_000_000
```
knighton, line 42: `predownload` is per worker btw (this seems large to me)

knighton:
> Q1: should we updt the dataloader to use [StreamingDataLoader](https://streaming.docs.mosaicml.com/en/stable/api_reference/generated/streaming.StreamingDataLoader.html) instead of the torch dataloader [here](https://github.com/mosaicml/examples/blob/abhi/streaming-023/common/text_data.py#L188)

The Composer trainer is integrated with DataLoader, uses StreamingDataset directory in resumption -- StreamingDataLoader is for non-Composer use cases

abhi-mosaic:
```
Q2: [Dataset timeout defaults to 120](https://github.com/mosaicml/examples/blob/abhi/streaming-023/common/text_data.py#L176); y not use 60 ie the [StreamingDataset default](https://streaming.docs.mosaicml.com/en/stable/api_reference/generated/streaming.StreamingDataset.html)?
```
I'll revert it to 60. I think in a previous version of Streaming we used 120 but perhaps the reasoning for that is outdated.

llm/yamls/mosaic_gpt/3b.yaml
```diff
@@ -73,7 +68,7 @@ scheduler:
 
 optimizer:
   name: decoupled_adamw
-  lr: 2.0e-4
+  lr: 1.6e-4
```
abhi-mosaic, line 71: @vchiley RIP

abhi-mosaic: I'm not sure if you were copying this exactly for the Legal LLM work but if so the LR may have been different from the GPT3 original hparams.

common/text_data.py
```diff
@@ -11,71 +11,76 @@
...
-        batch_size (Optional[int]): Hint batch_size that will be used on
-            each device's DataLoader. Default: ``None``.
+        remote (str, optional): Download shards from this remote path or directory. If None, this
+            rank and worker's partition of the dataset must all exist locally. Defaults to ``None``.
+        split (str, optional): Which dataset split to use, if any. Defaults to ``None``.
+        shuffle (bool): Whether to iterate over the samples in randomized order. Defaults to ``False``.
+        predownload (int, optional): Target number of samples ahead to download the shards of while
+            iterating. Defaults to ``100_000``.
+        keep_zip (bool, optional): Whether to keep or delete the compressed file when
+            decompressing downloaded shards. If set to None, keep iff remote is local. Defaults to
```
alextrott16, line 35: Typo: iff

PR #62, opened 2023-01-06
Fix bert check

dakinggg:
Discovered while factoring out things into a common module.

### DIFF
--- a/bert/src/bert_layers.py
+++ b/bert/src/bert_layers.py
@@ -101,7 +101,7 @@ class BertEmbeddings(nn.Module):
         inputs_embeds: Optional[torch.FloatTensor] = None,
         past_key_values_length: int = 0,
     ) -> torch.Tensor:
-        if (input_ids is not None) != (inputs_embeds is not None):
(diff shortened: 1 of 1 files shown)


bert/src/bert_layers.py
```diff
@@ -101,7 +101,7 @@ def forward(
         inputs_embeds: Optional[torch.FloatTensor] = None,
         past_key_values_length: int = 0,
     ) -> torch.Tensor:
-        if (input_ids is not None) != (inputs_embeds is not None):
+        if (input_ids is not None) == (inputs_embeds is not None):
```
alextrott16, line 104: I think `inputs_embeds` is usually left as `None`, so I'm wondering why I haven't seen this exception being raised... In any case, your fix looks correct. Assuming that the tester script runs successfully (confirm by running `composer main.py yamls/test/main.yaml`), then LGTM.

alextrott16 (approved):

bert/src/bert_layers.py
alextrott16: Got it. Thanks for catching this!

PR #63, opened 2023-01-06
Refactor

dakinggg:
This is the draft of factoring out a common module so people can see what it looks like and what they think. I started with refactoring common things out of BERT and LLM. The changes were already piling up, so vision folders can happen separately.

Note: It would probably be a good idea to make the repo pip installable so that we don't have to append to the path in order to have a common module.

Verification Todos:
- [x] run through the steps in LLM readme (convert c4, verify c4, start training run)
- [x] run through the steps in BERT readme (convert c4, verify c4, run the four test commands in the readme)
- [x] run LLM tests (pytest)


vchiley:
https://github.com/mosaicml/examples/pull/58 will integrate the triton kernel.
We should move that to the common folder as well.

alextrott16:
I'm generally in favor of this kind of refactor, but I think the open question about the overall structure (raised by @abhi-mosaic above) is important. Also, I really don't want us to end up re-building some evolution of YAHP.

My thinking is that we should not shy away from specific projects having lots of custom code when appropriate. This will give us a good chance to showcase examples that rely heavily on common utilities (e.g., BERT and GPT), as well as how to layer in customizations.

Using LegalGPT as an example, I think the decision of where to place that would depend on how much custom code it involves. If the only difference when using LegalGPT is changing the remote address of the data source, for example, then it would probably just make sense to be packaged into the GPT folder (because, in that case, it would very easily re-use the same `main.py`). However, if LegalGPT requires its own evaluators, metrics, etc. to the point that it would be cleaner to create a separate script for it, then I feel like it would deserve its own folder.

dakinggg:
@vchiley Yes, shared, complicated stuff like that should definitely be in common imo

@abhi-mosaic Yes, if it is not pip installable you would have to clone/fork/copy-paste in order to use `MosaicGPT` and `MosaicBERT`. My not totally thought through suggestion would be to follow the normal python repo structure, so have a single `mosaicml-examples` package. And have separate groups for each subfolder (e.g. `pip install mosaicml-examples[llm]`)

@abhi-mosaic / @alextrott16  I also agree it is unclear what to do about the folder structure. The current top level folders indicate model familys (llm), specific architectures (bert, resnet, deeplab), and datasets (cifar). It is difficult to balance the goals of organization and clean code with the goal of each example being self contained/easily copy pasteable. If we make the whole repo pip installable, it would allow legalgpt to `from examples.llm.mosaic_gpt import MosaicGPT`. I'm not sure I love this structure, but I think this is ok?. The legalgpt example is basically an application of mosaicgpt.

@alextrott16 Agree on not shying away from project having lots of specific code. I think there are basically two categories of stuff that I would like to see in common. 1) obvious shared code (e.g. `build_logger`). If you can easily use something from another example/common, do it/move it to common. 2) Shared things that are complex (e.g. flash attention/triton stuff). Having complex, error prone code in multiple places is something I would like to avoid. It is also confusing for a user if it is hard to immediately tell if there are differences between (e.g.) the `convert_c4` script in bert and the one in llm, and if those differences are salient. If a project needs something custom, it should just do it, and not try to shoehorn it into something more general.

dblalock (approved):
LGTM assuming tests + pre-commit checks are all happy.

dakinggg:
@dblalock I think there has already been drift away from precommit passing...so I only ran pre commit on the files that I modified

dakinggg:
@abhi-mosaic @alextrott16 how do you feel about merging this as is, as step one of refactoring? It is going to pretty painful to keep this PR up to date with other changes getting merged in.

alextrott16:
@dakinggg if tests are happy, then I'm fine with merging

abhi-mosaic:
I'm ok with merging this too. Will keep an eye as I will need to update #61 quite a bit, and I would really like to get that merged today/tomorrow.

PR #64, opened 2023-01-06
Docker build LLM workflow

bandish-shah:
# TL;DR 

Adds a basic GitHub Actions workflow to build Docker images for each example.  This PR builds an example image for the `llm` example only.  If image were generated based on the hash of the first commit in this PR, the following images would be pushed to Docker Hub:
* `mosaicml/examples:llm-latest`
* `mosaicml/examples:llm-242a6bd`

# Details

The workflow is triggered when:
* A pull request is created or updated
* The `main` branch is updated

When triggered by a pull requests, the workflow will only build a Docker image for each example specified in the workflow build matrix.  The images will not be pushed to the Docker Hub `mosaicml/examples` repository.

When triggered by a push to `main`, the workflow will build a Docker Image for each example specified in the workflow build matrix and push the images to the Docker Hub repository.  The images are tagged according to the following format in the `mosaicml/examples` repository:
* `mosaicml/examples:${{ matrix.example }}-latest`
* `mosaicml/examples:${{ matrix.example}-${{ github.sha }}`

To add examples, simply add the example folder name  to the list specified in the build matrix.

# To Do

- [x] Create `mosaicml/examples` Docker Hub repo: https://mosaicml.atlassian.net/browse/IT-1160
- [x] Remove `pull_request` trigger before merging

### DIFF
--- a/.github/workflows/build_docker.yml
+++ b/.github/workflows/build_docker.yml
@@ -39,7 +39,7 @@ jobs:
       - name: Build and Push the Docker Image
         uses: docker/build-push-action@v3
         with:
-          context: .
+          context: ./examples
           tags: mosaicml/examples:${{ matrix.example }}-latest,mosaicml/examples:${{ matrix.example }}-${{ env.IMAGE_TAG }}
           push: ${{ contains(github.ref, 'refs/heads/main') }}
           cache-from: type=registry,ref=mosaicml/examples:${{ matrix.example }}-buildcache
--- a/.github/workflows/build_docker.yml
+++ b/.github/workflows/build_docker.yml
@@ -14,15 +14,6 @@ jobs:
         example:
           - llm
     steps:
-
-      # - name: Maximize Build Space on Worker
-      #   uses: easimon/maximize-build-space@v6
-      #   with:
-      #     overprovision-lvm: true
-      #     remove-dotnet: true
-      #     remove-android: true
-      #     remove-haskell: true
-
       - name: Setup QEMU
         uses: docker/setup-qemu-action@v2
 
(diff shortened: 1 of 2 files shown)


bandish-shah:
@abhi-mosaic @samhavens this is a very basic workflow.  My goal was to have a single workflow that we can easily extend to all the examples in the repo but finer grained customizations per example are possible if needed.  Let me know what you think.

samhavens:
This is exactly what I was imagining when I made that Slack comment 🎉

abhi-mosaic:
@bandish-shah is there a lot of overhead / PR wait time if we create a tag on every commit? Like would it make sense to build these only on release tags of `mosaicml/examples`? I just don't know what the overhead is like.

If it's fast enough to do this on every commit, then I'm down to just merge this PR and try it out for a few weeks, see how the research team uses the new images :)

samhavens:
@abhi-mosaic since it's triggered on push to main, this should only happen on PR merge... but yeah, usually a PR is for one example, so this will trigger some redundant builds. But, it shouldn't block anything, as I don't think this will run before merging

abhi-mosaic (approved):
OK! Let's try it :)

bandish-shah:
@abhi-mosaic I just wanted to get something quick in place.  It shouldn't be a problem now since this is just LLM for now.  Also once the Docker images are built they get cached so if there are no changes it _should_ be pretty quick.  Right now it looks like no other CI runs in this repo so this should take more than a couple of minutes to run.

As we add more examples, we should add path based filtering so only the builds for examples that have changed run.  It's not straight forward to do with a matrix build but that is the next major enhancement to this flow.  Worst case brute force is to just have an independent build workflow for each example that triggers based on path.

bandish-shah:
@samhavens it does currently run on `pull_request` events for development purposes but let me add a TODO to remove that before the final merge so it only runs on `push` events to the `main` branch.

PR #65, opened 2023-01-10
Disable Docker GHA job for forked repos

bandish-shah:
Currently the Docker GHA flow will run for pushes to `main` branch in the base and forked repos.  This PR gates the Docker job with a repo owner check and will only run if the owner is `mosaicml`.

The Docker GHA workflow is not compatible with forks since it requires a repo secret to run.


vchiley (approved):
lgtm

PR #66, opened 2023-01-11
Cleanup Docker image

abhi-mosaic:
I am not a Github-Actions expert, but I tried editing the Dockerfile so that it just copies, installs, and deletes the `${EXAMPLE}/requirements.txt` file.


bandish-shah (approved):
LGTM

PR #67, opened 2023-01-11
Run composer ICL eval w/ MosaicGPT

bmosaicml:
This PR implements functionality to run a composer-based ICL eval for hugging face models of mosaic gpt checkpoints. This implementation is much faster than the previous eleuther-based implementation and should be the default going forward.

This PR does not implement functionality to log ICL task results during training. I will add that in a later PR.


llm/icl_eval/model_loading.py
```diff
@@ -0,0 +1,118 @@
...
+    parsed_path = urlparse(path_to_download)
+    download_path = (parsed_path.path if parsed_path.scheme == 's3' else
+                     parsed_first_checkpoint.path).lstrip('/')
+    checkpoint_name = _get_checkpoint_name_from_path(parsed_path.path)
+    local_path = os.path.join(CHECKPOINT_DIR, checkpoint_name)
+    if not os.path.exists(local_path):
+        get_file(destination=local_path,
+                 path=download_path,
+                 object_store=load_object_store,
+                 progress_bar=True)
```
eracah, line 49: This duplicates a lot of logic already in Composer. I'm not saying this isn't the right way to go b/c Composer checkpointing interface/ease of use needs improvement. I'm just curious what pain points led you to write this function, etc, so we can improve our checkpointing interface

llm/icl_eval/model_loading.py
```diff
@@ -0,0 +1,119 @@
...
```
eracah, line 47: You actually don't need to make an object store yourself. You can just pass in the s3 URI
```suggestion
    parsed_path = urlparse(path_to_download)
    checkpoint_name = _get_checkpoint_name_from_path(parsed_path.path)
    local_path = os.path.join(CHECKPOINT_DIR, checkpoint_name)
    if not os.path.exists(local_path):
        get_file(destination=local_path,
                 path=path_to_download,
                 progress_bar=True)
```

eracah (changes requested):

llm/icl_eval/model_loading.py
```diff
@@ -0,0 +1,119 @@
...
+
+    return checkpoint_name
+
+
+def init_huggingface_causal_lm(
+        config: str
+) -> Dict[str, Union[AutoModelForCausalLM, AutoTokenizer]]:
+    return  create_gpt2(use_pretrained=True, pretrained_model_name=config)
```
abhi-mosaic, line 55: I don't think we should be using this method, might be related to our OOM issues

llm/icl_eval/evaluate_model.py
```diff
@@ -0,0 +1,86 @@
...
+    with open(yaml_path) as f:
+        yaml_cfg = om.load(f)
+    cli_cfg = om.from_cli(args_list)
+    cfg = om.merge(yaml_cfg, cli_cfg)
+
+    model = load_model(**cfg.get("model"))
+    evaluators, logger_keys = build_evaluators(cfg)
+    in_memory_logger = InMemoryLogger()  # track the logged metrics in the in_memory_logger
+   
+    trainer = Trainer(model=model, max_duration='1ba', loggers=in_memory_logger)
```
abhi-mosaic, line 74: Was `max_duration` a requirement from the Trainer? That's a UX bug we should report to Composer team if so, because we are only trying to use `trainer.eval()`

llm/mcloud/mcli-gpt-neo-eval.yaml
```diff
@@ -0,0 +1,66 @@
...
+integrations:
+  - integration_type: git_repo
+    git_repo: mosaicml/examples
+    git_branch: feature/composer_icl_eval
+    pip_install: -r llm/requirements.txt
+
+# We are fetching, converting, and training on the 'val' split
+# as it is small and quick to get going for this demo.
+# For real training runs, follow the instructions in `examples/llm/README.md`
+# to convert and host the full 'train' dataset.
```
abhi-mosaic, line 14: remove

PR #70, opened 2023-01-12
Update YAMLS

abhi-mosaic:
- Turns off `progress_bar` by default to improve UX when using MCLOUD
- Update `llm/mcloud` examples to new format

### DIFF
--- a/llm/mcloud/mcli-1b-custom.yaml
+++ b/llm/mcloud/mcli-1b-custom.yaml
@@ -1,8 +1,8 @@
 integrations:
-  - integration_type: git_repo
-    git_repo: mosaicml/examples
-    git_branch: v0.0.2
-    pip_install: -r llm/requirements.txt
+- integration_type: git_repo
+  git_repo: mosaicml/examples
+  git_branch: v0.0.2
+  pip_install: -r llm/requirements.txt
 
 # We are fetching, converting, and training on the 'val' split
 # as it is small and quick to get going for this demo.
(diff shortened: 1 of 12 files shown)


llm/mcloud/mcli-1b-custom.yaml
```diff
@@ -26,19 +26,20 @@ cluster: r0z0 # replace with your cluster here!
 # and referenced in our `command` above
 # We use this to customize hparams like `max_seq_len=8192` at runtime
 parameters:
-  data_remote: &data_remote ./my-copy-c4
   data_local: &data_local ./my-copy-c4
-  max_seq_len: &max_seq_len 8192
```
bmosaicml, line 31: Can you explain these changes in the PR?

bmosaicml (approved):

llm/mcloud/mcli-1b-custom.yaml
```diff
@@ -26,19 +26,20 @@ cluster: r0z0 # replace with your cluster here!
...
   model:
     name: mosaic_gpt
     device: meta
     tokenizer_name: *tokenizer_name
-    d_model: 2048
-    n_heads: 16 # Modified 24->16 so that d_head == 128 to statisfy FlashAttention
-    n_layers: 24
+    d_model: 768
+    n_heads: 12
+    n_layers: 12
```
vchiley, line 42: This makes it a 125M model, not a 1B model.

llm/mcloud/mcli-1b-custom.yaml
abhi-mosaic: Ah, I think as I was trying to update it, I copied over the 125M model and did not preserve some of the custom flags. Fixing it now!

abhi-mosaic:
Reverted the `mcloud/` folder as it should not actually be changed until we have a new stable pin, and then those examples should point at the stable pin.

PR #71, opened 2023-01-13
Fix SpeedMonitorMFU + DDP

abhi-mosaic:
Composer wraps the `state.model` in a `DistributedDataParallel` object if using DDP. The `SpeedMonitorMFU` wasn't handling this case and failed to log MFU.

This PR fixes that, and also adds some guard rails for DDP in the `main.py` training script.

This is the message that `main.py` prints out if you attempt to use `meta` init device with DDP:
```
Using init device `cfg.model.device='meta'` is only valid when using FSDP! Reverting to `cfg.model.device='cpu'`.
```

Here is MFU for DDP vs. FSDP:
Screenshot 2023-01-12 at 4 34 33 PM

### DIFF
--- a/common/speed_monitor_w_mfu.py
+++ b/common/speed_monitor_w_mfu.py
@@ -14,6 +14,7 @@ from typing import Any, Deque, Dict, Union
 import torch
 from composer.core import Callback, State
 from composer.loggers import Logger
+from composer.models.base import ComposerModel
 from composer.utils import dist
 
 GPU_AVAILABLE_FLOPS = {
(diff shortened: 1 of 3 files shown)


common/speed_monitor_w_mfu.py
```diff
@@ -200,8 +201,11 @@ def batch_end(self, state: State, logger: Logger):
                 logger.log_metrics({'throughput/tokens_per_sec': samples_per_sec * state.dataloader.dataset.max_seq_len})
                 logger.log_metrics({'throughput/device/tokens_per_sec': dev_samples_per_sec * state.dataloader.dataset.max_seq_len})
 
-            if hasattr(state.model, 'num_fwd_flops'):
-                flops_per_sec = (3 * state.model.num_fwd_flops) * samples_per_sec
+            composer_model = state.model
+            if not isinstance(composer_model, ComposerModel):
+                composer_model = composer_model.module # Handles DDP case until Composer fixes this
+            if hasattr(composer_model, 'num_fwd_flops'):
+                flops_per_sec = (3 * composer_model.num_fwd_flops) * samples_per_sec
```
bmosaicml, line 208: Can you add comment explaining the constant 3?

bmosaicml (approved):

PR #72, opened 2023-01-13
Refactor C4 conversion script

abhi-mosaic:
This PR fixes some bugs, and adds a `train_small` split that can be used for demos rather than point the train_loader at `val`. The latter style has been noted as a potential bug by `mosaicml-streaming` team, which they are working on, but we should probably follow the safer convention either way.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -67,11 +67,11 @@ repos:
   hooks:
   - id: pyright
     name: pyright
-    entry: pyright .
+    entry: pyright
     language: node
     types: [python]
-    pass_filenames: false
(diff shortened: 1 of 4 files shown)


bmosaicml (approved):
It looks good to me but can you describe how you tested it? Also reproduce the readme instruction and ensure they work properly?

abhi-mosaic:
Ah so for testing, I copy-pasted the README instructions to confirm that it works. Let me see if I can add/modify some existing pytests too to make this automated... that's a good idea.

abhi-mosaic:
Deferring to a followup PR.

PR #74, opened 2023-01-17
Remove codeowners file

dblalock:
Based on slack discussion, having mandatory codeowner checks is more trouble than it's worth. I tried making them optional in the repo settings a few days ago but it seems to still add everyone to the PR as long as the CODEOWNERS file exists.

So this PR just removes the CODEOWNERS file.


abhi-mosaic (approved):
LGTM!

PR #75, opened 2023-01-17
Fix precommit checks

dblalock:
Trying to get everything in a passing state so my precommit CI PR can be minimal.

All the changes are autoformatting except:

- A couple manual pydocstyle fixes in llm/src
- Some manual `type: ignore`s in llm/main.py
- Some manual formatting in a few yaml files


abhi-mosaic (approved):
LGTM!

PR #76, opened 2023-01-19
Make repo pip installable

dblalock:
Addresses [#319](https://mosaicml.atlassian.net/browse/RESEARCH-319). This makes it easier to have common code for different subdirectories, as well as to run the linting.

The non-obvious choice here is that I renamed `common` to `mosaic_examples`. This way the package name that gets installed makes sense to users.

Main thing to review for is whether this subtlely broke the imports that `sys.path.append` was handling before.

Update based on what @abhi-mosaic wanted to try:
Got it training a GPT-125m with this MCLI yaml:
```yaml
integrations:
- integration_type: git_repo
  git_repo: mosaicml/examples
  git_branch: davis/pip-installable

# We are fetching, converting, and training on the 'val' split
# as it is small and quick to get going for this demo.
# For real training runs, follow the instructions in `examples/llm/README.md`
# to convert and host the full 'train' dataset.
command: |
  cd examples
  pip install .[llm]
  python scripts/convert_c4.py --out_root llm/my-copy-c4 --splits train_small val
  cd llm
  composer main.py yamls/mosaic_gpt/125m.yaml \
    train_loader.dataset.split=train_small \
    run_name=$COMPOSER_RUN_NAME
image: mosaicml/pytorch:1.13.1_cu117-python3.10-ubuntu20.04
optimization_level: 0

run_name: mosaic-gpt-125m-gpus-8

gpu_num: 8
gpu_type: a100_40gb
cluster: r7z2 # replace with your cluster here!

# The below is injected as a YAML file: /mnt/config/parameters.yaml
# but is not used in this example.
parameters: {}

```

Although it eventually crashed with this error, though I assume it's unrelated:
```
...
[batch=273/4800]:
	 Train LanguageCrossEntropy: 5.7413
	 Train Perplexity: 311.4801
	 Train loss/train/total: 5.7413
[E ProcessGroupNCCL.cpp:821] [Rank 7] Watchdog caught collective operation timeout: WorkNCCL(SeqNum=24064, OpType=_ALLGATHER_BASE, Timeout(ms)=1800000) ran for 1800367 milliseconds before timing out.
[E ProcessGroupNCCL.cpp:456] Some NCCL operations have failed or timed out. Due to the asynchronous nature of CUDA kernels, subsequent GPU operations might run on corrupted/incomplete data.
[E ProcessGroupNCCL.cpp:461] To avoid data inconsistency, we are taking the entire process down.
terminate called after throwing an instance of 'std::runtime_error'
  what():  [Rank 7] Watchdog caught collective operation timeout: WorkNCCL(SeqNum=24064, OpType=_ALLGATHER_BASE, Timeout(ms)=1800000) ran for 1800367 milliseconds before timing out.
```


abhi-mosaic (approved):
Ran the YAML on my side and verifies that it works! Approving to unblock.

abhi-mosaic (approved):
Unblocking, as the bug seems to be unrelated to this PR. Sorry for delay @dblalock !

PR #77, opened 2023-01-20
Fix new pyright errors

dblalock:
We keep pyright errors added, so this fixes the new ones since I last fixed them all. (Protip: run `make lint` to check for these errors).

Nothing about the program semantics should have changed except that we no longer pass in args that were removed from composer (see [this issue](https://mosaicml.atlassian.net/browse/RESEARCH-364) and associated slack post).

With everything passing, I'll (in theory) get the pre-commit checks added to the CI tomorrow so we stop adding these errors.


llm/yamls/mosaic_gpt/fast.yaml
```diff
@@ -0,0 +1,29 @@
+integrations:
```
abhi-mosaic, line 1: This file should probably be uncommitted right?

abhi-mosaic (approved):
LGTM except for that extra yaml file`fast.yaml`.

PR #79, opened 2023-01-20
Add first round of CI checks

dblalock:
Just adds linting checks, since it already took quite a bit to get those working. See bottom of this PR for what the checks will look like.


llm/requirements.txt
```diff
@@ -1,14 +1,16 @@
...
+einops==0.5.0
 mosaicml==0.12.0
 mosaicml-streaming==0.2.2
 flash-attn==0.2.2
+mosaicml-cli>=0.2.32,<1
 # need a newer version of triton
 triton==2.0.0.dev20221103
 transformers==4.24.0
-datasets==2.7.1
+datasets>=1.15.1,<3  # low version to avoid conflict with lm_eval
```
abhi-mosaic, line 12: This will break, I don't think we can allow past datasets 2.7.1

abhi-mosaic: Prefer <=2.7.1

abhi-mosaic (approved):

PR #81, opened 2023-01-22
need to override split twice in CLI configs

galtay:
I think `cli_cfg` needs to come first in order to override `yaml_cfg` (or at least it did on my system)

### DIFF
--- a/llm/main.py
+++ b/llm/main.py
@@ -176,5 +176,5 @@ if __name__ == '__main__':
     with open(yaml_path) as f:
         yaml_cfg = om.load(f)
     cli_cfg = om.from_cli(args_list)
-    cfg = om.merge(yaml_cfg, cli_cfg)
+    cfg = om.merge(cli_cfg, yaml_cfg)
     main(cfg)


galtay:
... on second glance, the order doesn't matter ... its just that there seem to be two things to override (if following the README) 

```
composer main.py yamls/mosaic_gpt/125m.yaml train_loader.split=train_small
```

results in 

```
'train_loader': {'name': 'text', 'dataset': {'local': './my-copy-c4', 'remote': None, 'split': 'train', 'shuffle': True, 'tokenizer_name': 'gpt2', 'max_seq_len': 2048, 'group_method': 'concat', 'shuffle_seed': 17}, 'drop_last': True, 'num_workers': 8, 'split': 'train_small'}
```

so the command to override the training split should be something like, 
```
composer main.py yamls/mosaic_gpt/125m.yaml train_loader.split=train_small train_loader.dataset.split=train_small
```

abhi-mosaic:
Hi @galtay, thanks for the report! I think this is just a mistake in our README.md, you should only have to change `train_loader.dataset.split`. Will fix!

abhi-mosaic:
Closing this as the README is fixed in #82 .

We will also look into using OmegaConf syntax to bind `cfg.remote` in a way that is more user friendly.

PR #83, opened 2023-01-24
HF FSDP wrap BLOOM and OPT as well

samhavens:
This makes the FSDP wrap function a bit more generic so it will work on OPT and BLOOM models.

However, it removes an assert, so the user could now call it on any model — we may want some sanity check that they aren't calling it on e.g. T5, which this doesn't support yet (would need to look at `model.encoder` for some properties, I think). For now, I raise a ValueError if any of the expected parts of the model aren't found.

Also, I wasn't sure on how to correctly attribute the helper functions, see the file header.


abhi-mosaic:
Looking good so far! I think we should also consider using Composer's `HuggingFaceModel` class and see if it works/breaks. It gives us a lot of good functionality for free.

samhavens:
Following up: this code works to FSDP-wrap OPT and BLOOM. I tested on 4x A100 40G, and OPT 6.7B and BLOOM 3B works

alextrott16:
Love this. Simple and easy to follow/extend.

It would be good to confirm that it works on T5 (an Encoder-Decoder) and BERT (an Encoder-only). Along those lines, I might recommend explicitly defining the model types that this supports (with unsupported models triggering a warning or error) since there are a _lot_ of HF model types out there and the 10 most popular ones will probably 99% of use cases.

llm/src/hf_causal_lm.py
```diff
@@ -1,45 +1,137 @@
...
-    # When using the HF Causal LM models,
+    """
+    Wrap any model for FSDP which follows one of the 3 existing conventions from
+    HuggingFace for decoder-only LLMs
+    """
+    causal_base_model = hf_get_causal_base_model(model)
+    model_block = hf_get_causal_hidden_layers(model)[0]
+    block_type = type(model_block)
+    lm_head = hf_get_causal_hidden_layers(model)
+    tied_embeddings = hf_get_tied_embedding_weights(causal_base_model)
```
samhavens, line 119: If any of these are None should we warn or error? @alextrott16

samhavens:
@alextrott16 I agree that we should give information about confidence. There are 2 levels that I see: "we haven't checked this model yet, but it has all the properties that we expect in the helper functions" (likely will work?) vs "we tried to wrap this model, and were unable to find needed components" (almost certainly won't work)

samhavens:
Are the `hf_causal_lm` examples currently working?

samhavens:
> @alextrott16 I agree that we should give information about confidence. There are 2 levels that I see: "we haven't checked this model yet, but it has all the properties that we expect in the helper functions" (likely will work?) vs "we tried to wrap this model, and were unable to find needed components" (almost certainly won't work)

I've implemented the later as a `ValueError`. Now need to test BERT-style & T5-style

llm/src/hf_causal_lm.py
```diff
@@ -1,85 +1,187 @@
...
-        module, _HF_MODEL_BLOCKS)
+        module, block_type)
 
 
-class ComposerHFCausalLM(ComposerModel):
-
-    def __init__(self, cfg):
-        super().__init__()
+class ComposerHFCausalLM(HuggingFaceModel):
+    def __init__(self, cfg: DictConfig):
```
abhi-mosaic, line 151: Looking at the flow here, I think we are taking a `cfg`, creating some HF objects, then initializing a `HuggingFaceModel`.  Could this be turned into a function rather than a subclass? Basically, this would demonstrate that `HuggingFaceModel` is usable without requiring writing your own ComposerModel class.

Also, if this is not possible, then we will have some feedback for Composer team :)

.gitignore
```diff
@@ -1,3 +1,6 @@
+# Data
+my-copy-c4/
```
abhi-mosaic, line 2: <3

llm/src/hf_causal_lm.py
samhavens: So I tried that, and that is how I ended up with those mis-shapen tensor errors. I had to override `loss`, `update_metric`, like was being done before in the class that wrapped `ComposerModel`.

dakinggg: I think once I merge https://github.com/mosaicml/composer/pull/1857 we might be able to do this. (I think) RN the reshaping is necessary because you are using `Perplexity` which uses `HFCrossEntropy` which doesn't reshape for you.

dakinggg: there might still remain a difference in the loss bit (our `HuggingFaceModel` assumes that the underlying HF model returns the loss as part of its output), but i'd have to look into our implementation to check that.

abhi-mosaic: Thanks @dakinggg ! I think in that case, we can merge as is for the functionality, with a TODO to cleanup in the future, so that we can use `HuggingFaceModel` directly.

samhavens:
@alextrott16 RE: model incompatibility concerns: 

* this seems to wrap Codegen (a decoder-only model I hadn't looked at). 
* the codebase errors if you try to pass T5, so didn't check that one (we call AutoModelForCausalLM, which substantially narrows down the potential models)
* it raises a ValueError if you try to use Bert

So I think we are good on that front!

abhi-mosaic (approved):
Thanks for untangling FSDP and expanding our supported models!

PR #89, opened 2023-01-27
Fix make test, update readme with test + lint instructions

dblalock:
In theory this just updated the main readme, but then I realized that `make test` no longer calls the right script. Also, all the subdirectories except bert should have a `cd` command in their fork/installation instructions.


abhi-mosaic (approved):
LGTM

PR #90, opened 2023-01-27
Make everything pip installable

abhi-mosaic:
This PR touches almost every file, but it is doing very little real editing, and is mainly shuffling files around:
* move all code under `examples/examples`
* change `setup.py` to install everything
* update READMEs and import statements


examples/llm/throughput/run_all_configs.py
```diff
@@ -257,13 +257,13 @@ def run_config(config, args, project, image, RUN):
     # Define our command
     if streaming_data:
         command = """
-        composer examples/llm/main.py /mnt/config/parameters.yaml
+        composer examples/common/llm/main.py /mnt/config/parameters.yaml
```
dblalock, line 260: wrong dir?

```suggestion
        composer examples/examples/llm/main.py /mnt/config/parameters.yaml
```

examples/llm/throughput/run_all_configs.py
```diff
@@ -257,13 +257,13 @@ def run_config(config, args, project, image, RUN):
...
         """
     else:
         command = """
-        python examples/llm/convert_c4.py --out_root ./my-copy-c4 --splits val
+        python examples/common/convert_c4.py --out_root ./my-copy-c4 --splits val
 
-        composer examples/llm/main.py /mnt/config/parameters.yaml
+        composer examples/common/llm/main.py /mnt/config/parameters.yaml
```
dblalock, line 266: also wrong dir?
```suggestion
        composer examples/examples/llm/main.py /mnt/config/parameters.yaml
```

examples/llm/train.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+source /workdisk/abhi/examples/venv-new/bin/activate
+composer main.py yamls/mosaic_gpt/125m.yaml train_loader.dataset.split=train_small
```
dblalock, line 3: Is this file supposed to be here? It doesn't seem to be referenced anywhere and assumes there's a `venv-new` that something else creates + cleans up

examples/requirements.txt
```diff
@@ -0,0 +1,16 @@
+# These requirements are shared for all examples
+# Each example subdir has its own `requirements.txt` with extra deps.
```
dblalock, line 2: is there a reason the repo-wide requirements aren't in the repo rootdir? I would expect at least the dev requirements (pre-commit, pytest, etc) to not be under examples.

setup.py
```diff
@@ -5,19 +5,17 @@
...
 
 _PACKAGE_NAME = 'mosaicml-examples'
-_PACKAGE_DIR = 'mosaicml_examples'
-_EXAMPLE_SUBDIRS = ('cifar', 'resnet', 'deeplab', 'bert', 'llm')
-
-_project_root = os.path.dirname(os.path.realpath(__file__))
-
-print('_project_root:', _project_root)
+_EXAMPLE_SUBDIRS = ['cifar', 'resnet', 'deeplab', 'bert', 'llm']
+_PACKAGES = find_packages(exclude=['*tests*', '*scripts*'])
```
dblalock, line 12: Is the goal still to just have `examples` be the one package? I think this will make any subdir with an __init__.py a package, which might be unintended behavior down the road.

examples/resnet/tests/__init__.py
```diff
@@ -0,0 +1,2 @@
+# Copyright 2022 MosaicML Examples authors
+# SPDX-License-Identifier: Apache-2.0
```
dblalock, line 2: do we want the tests to be a package? If so we should make sure to do this for all test subdirs.

I kinda don't want it to be a package unless there's code sharing across test files so that we have to explicitly opt into importing stuff from test files. Just part of a general defensive policy wrt enforcing modularity.

dblalock (commented):
LGTM aside from local nitpicks. The only part that seems weird to me is moving requirements.txt out of the repo root.

examples/bert/yamls/main/mcloud_run.yaml
```diff
@@ -100,26 +102,27 @@ parameters:
...
+#   (Optional) W&B logging
+#   loggers:
+#     wandb:
+#       project:
+#       entity:
+
+#   (Optional) Checkpoint to local filesystem or remote object store
+#   save_interval: 3500ba
+#   save_num_checkpoints_to_keep: 1  # Important, this cleans up checkpoints saved to DISK
+#  save_folder:      # e.g. './{run_name}/ckpt' (local) or 's3://mybucket/mydir/{run_name}/ckpt' (remote)
```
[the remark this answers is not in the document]
abhi-mosaic: Yup, disabling checkpointing by default.

examples/llm/train.sh
abhi-mosaic: Yup this is my bad, will remove.

examples/requirements.txt
abhi-mosaic: Will move the `requirements.txt` to the top level folder

dblalock (approved):
LGTM except that the Makefile also a line for BERT under the `test` target

PR #92, opened 2023-01-27
Add end-to-end tests for main.py and glue.py

alextrott16:


examples/bert/tests/test_glue.py
```diff
@@ -0,0 +1,34 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+import pytest
+import shutil
+import tempfile
+from typing import Any
+
+from omegaconf import OmegaConf
+
+from ..glue import train
```
abhi-mosaic, line 11:
```suggestion
from examples.bert.glue import train
```

examples/bert/tests/test_glue.py
```diff
@@ -0,0 +1,34 @@
...
+        return self.path
+
+    def __exit__(self, exc_type: Any, exc_value: Any, traceback: Any):
+        del exc_type, exc_value, traceback # unused
+        shutil.rmtree(self.path)
+
+@pytest.mark.parametrize('model_name', ['mosaic_bert', 'hf_bert'])
+def test_glue_script(model_name: str):
+
+    with open('tests/smoketest_config_glue.yaml') as f:
```
abhi-mosaic, line 26: I'm not an expert at path stuff, but wondering if this requires the `pytest tests` be run from only the directory `examples/bert`? Would CI have trouble if running pytests from farther away?

abhi-mosaic (commented):
Made some quick comments! I think to get tests working with `make` and CI, you'll also want to edit the file `test.yaml`, and you'll see a line to uncomment

examples/bert/tests/test_glue.py
alextrott16: It might have trouble, but this is the pattern set across some of the other examples (e.g. cifar) when a "smoketest" config is part of the local `tests/`

alextrott16: Taking a closer look at `scripts/test_subdirectory.py`, it look like the steps are to `cd` into the given subdirectory and run `python -m pytest tests`, so the way I have it set up should work :)

abhi-mosaic: Sweet! Thanks for explaining :)

abhi-mosaic (approved):
LGTM! Sadly we won't get have CI testing these edits until we merge into the the other branch, but assuming this passed the pre-commit it's probably relatively safe.

PR #93, opened 2023-01-27
add cost estimator

vchiley:
add cost estimate to speed monitor

eg output
Screenshot 2023-01-27 at 3 51 19 PM
ie it'll cost about ~$0.86 to train GPT125M on 200 batches of data on a 80GB A100 (assuming the price of that GPU is $2.75).
[w&b](https://wandb.ai/mosaic-ml/cost_estimate_test)


abhi-mosaic:
Couple high level comments:

* this is a plot of the live estimate of the whole run (stateless, based on current throughput)
* ... could we also get a plot of the $ spent so far (monotonic increasing line). This metric could then be used as an x-axis for other plots :)
* to make the second plot elastic-resilient, we need to be keeping track of something like `self.device_time += n_devices * dt`, and storing it in the callback state. Then upon resumption, the line will be consistent, just like the `wall_clock/...` plots.
* we should preemptively search for an env var that MCLOUD can inject in the future. Maybe something like `DEVICE_PRICE_PER_HOUR`. 
* Probably a good idea to sync with @mvpatel2000 who I think is working on something very similar for Composer

mosaicml_examples/builders.py
```diff
@@ -25,7 +25,9 @@ def build_callback(name, kwargs):
             return SpeedMonitor(window_size=kwargs.get('window_size', 1))
         return SpeedMonitorMFU(window_size=kwargs.get('window_size', 1),
                                gpu_flops_available=kwargs.get(
-                                   'gpu_flops_available', None))
+                                   'gpu_flops_available', None),
+                               gpu_cost_per_hour=kwargs.get(
+                                   'gpu_cost_per_hour', None))
```
abhi-mosaic, line 30: Something like this to future proof? ;)

```suggestion
                               gpu_cost_per_hour=kwargs.get(
                                   'gpu_cost_per_hour', os.environ.get('DEVICE_PRICE_PER_HOUR', None)))
```

mvpatel2000:
@vchiley lets talk offline on this

mosaicml_examples/builders.py
vchiley: I moved this to the calculator instead of getting it in the builder.

vchiley:
updated to include running cost based on running device time
Screenshot 2023-02-07 at 11 38 20 PM
Screenshot 2023-02-07 at 11 39 25 PM

mvpatel2000 (commented):
I would recommend not including total cost for now. Estimating total time is pretty complicated, especially for vision (eg the code would break I think if u ran w max dur = epochs). If you'd like to include it, lets chat offline

vchiley:
mihir and i decided offline that having the cost estimate is fine, even tough it will NOT be a perfect estimate.

dakinggg:
Closing, no longer belongs on this repo.

PR #94, opened 2023-01-28
Allow dictconf to be null

dskhudia:
config for loggers, callbacks and algorithms can now be overridden via commandline. 

Example: 
`composer main.py yamls/mosaic_gpt/125m.yaml algorithms=null`

### DIFF
--- a/examples/llm/main.py
+++ b/examples/llm/main.py
@@ -114,22 +114,24 @@ def main(cfg):
     # Scheduler
     scheduler = build_scheduler(cfg.scheduler)
 
+    # we use (cfg.get(...) or {}) instead of cfg.get(..., {}) so that
+    # .items() works even when the value is None
     # Loggers
     loggers = [
         build_logger(name, logger_cfg)
-        for name, logger_cfg in cfg.get('loggers', {}).items()
+        for name, logger_cfg in cfg.get('loggers' or {}).items()
     ]
 
     # Callbacks
     callbacks = [
         build_callback(name, callback_cfg)
-        for name, callback_cfg in cfg.get('callbacks', {}).items()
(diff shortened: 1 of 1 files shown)


dblalock:
when you say "can be overridden," do you mean "can be set to None" specifically? And did this not work before? I didn't try it but I would expect `[whatever(k, v) for k, v in cfg.get(maybe_dict, {}]` to result in `[]` if maybe_dict is `None`. Could just be me not understanding some details about omegaconf though.

dskhudia:
It didn't work before. with `composer main.py yamls/mosaic_gpt/125m.yaml algorithms=null` `cfg.get('algorithms', {})` results in `None` and that has no `items` method.

dblalock (approved):
Questions resolved via slack discussion.

My main concern is that readers (such as apparently me after a long day...) might not think about the edge case (i.e., `cfg.get('key', {})` returning None if `key`'s value is None, rather than missing) and be tempted to shorten this code, silently breaking stuff. 

To help mitigate this risk (and save space in general), I suggested a more compact version + added a quick mention of why it's written this way.

dskhudia:
@dblalock Thanks for the suggestions. Accepted and will merge if the tests pass.

abhi-mosaic:
@dskhudia heads up, I think we need to rebase this change on `main` now that #90 has been merged.

dblalock:
So I think we accidentally merged `(cfg.get('algorithms' or {})` instead of `(cfg.get('algorithms') or {})`. Which probably won't work?

dskhudia:
uh-oh! Let me make fix PR. I am surprised CI tests didn't detect it.

dskhudia:
@dblalock added https://github.com/mosaicml/examples/pull/101

PR #95, opened 2023-01-28
Triton kernel alibi

vchiley:
POC; NOT tested; it does run in MosaicGPT; loss starts going down but no convergence runs yet
```
        if ALIBI:
            alibi_bias_max = 8
            alibi = (offs_m[:, None] - seqlen_q + 1) - ((start_n + offs_n)[None, :] - seqlen_k + 1)
            alibi = -1 * tl.abs(alibi)
            m = (off_h + 1) * (alibi_bias_max / nheads)
            alibi = alibi / pow(2, m)
            qk += alibi
```

The above adds alibi to the triton kernel itself so we don't need to take care of it as a mask on the torch side.
Should be slightly faster and should result in faster code since we do not need to load the alibi mask at every kernel launch.

Proposal for similar additions: 
- We can potentially also pass `key_padding_mask` to the kernel
- We can also probably add attn masking for concatenated seq into the kernel itself (using some special concat token or something)

### DIFF
--- a/llm/src/flash_attention.py
+++ b/llm/src/flash_attention.py
@@ -40,6 +40,7 @@ class FlashAttention(nn.Module):
             key_padding_mask: Optional[Tensor] = None,
             attn_mask: Optional[Tensor] = None,
             is_causal: bool = False,
+            alibi: bool = False,
             need_weights: bool = False,
             average_attn_weights: bool = True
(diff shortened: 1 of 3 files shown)


vchiley:
Given we're just using the triton kernel from FlashAttn repo we'll close this for now.

PR #97, opened 2023-01-30
Register `low_precision_layernorm` algorithm

abhi-mosaic:
Adding this here so we can diff against `fused_layernorm`, and if successful, replace the two inside the `bert` example.

### DIFF
--- a/examples/common/builders.py
+++ b/examples/common/builders.py
@@ -3,7 +3,8 @@
 
 import composer
 from composer import algorithms
-from composer.callbacks import LRMonitor, MemoryMonitor, SpeedMonitor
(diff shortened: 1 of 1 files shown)


alextrott16 (approved):
LGTM. In a perfect world, we would catch the case where someone uses both `fused_layernorm` and `low_precision_layernorm`, but that's not essential because it probably won't come up.

abhi-mosaic:
> LGTM. In a perfect world, we would catch the case where someone uses both fused_layernorm and low_precision_layernorm, but that's not essential because it probably won't come up.

Agreed, but I wonder if this is belongs as a Composer-level safety check. I'll make a JIRA for it.

PR #99, opened 2023-01-30
Add UL2 code to examples

alextrott16:
WIP. This PR introduces UL2 to the examples repo, which features a new data collator for mixture-of-denoisers training, support for HuggingFace T5 and Prefix LM (converted Causal LM) models, among other utilities.

TODO:

- [ ] Top-level README
- [ ] SuperGLUE README
- [x] SuperGLUE support for decoder-only models
- [ ] tests
- [x] FSDP support
- [ ] Missing docstrings


dblalock:
So I see this WIP but also that you requested review...is there some aspect you're hoping to get feedback on?

alextrott16:
@dblalock Sorry for not adding any context to the review request. At this point, any feedback on high-level structure (including whether this should be folded into LLM) would be useful. If you feel so motivated as to take a deeper look, any resulting feedback would also be very appreciated.

I don't think there's anything urgent, though. So, it's also perfectly fine with me if you want to ignore this PR until I have more specific review requests.

examples/ul2/src/mod_print_callback.py
```diff
@@ -0,0 +1,214 @@
...
+        """Verifies model class."""
+        del logger  # unused
+
+        assert hasattr(
+            state.model, 'model'
+        ), 'This callback requires that state.model is a HuggingFaceWrapper instance.'
+        hf_model = state.model.model
+        assert isinstance(
+            hf_model, PreTrainedModel
+        ), 'This class only works for models that subclass `transformers.PreTrainedModel`.'
```
dblalock, line 63: we should raise Exception subclasses rather than `assert`-ing when we think the case might actually happen and we aren't in a test. Maybe a TypeError here?

Also, it would be better to [test for behavior, not implementation](https://testing.googleblog.com/2013/08/testing-on-toilet-test-behavior-not.html). What interface do we need from these objects? Whatever it is, we should ideally just test for that at the point where it's used.

Though if we require a huge interface that only this class would ever implement, or actually _are_ leaning on implementation details, that could be an exception. (These both suggest some pretty tight coupling though...)

examples/ul2/src/mod_print_callback.py
```diff
@@ -0,0 +1,214 @@
...
+        """Periodically prints formatted outputs generated by the model."""
+        del logger  # unused
+
+        if dist.get_global_rank() != 0:
+            return
+
+        if not self._verified_collator:
+            assert isinstance(
+                state.train_dataloader.collate_fn, MixtureOfDenoisersCollator
+            ), 'The dataloader must use the `MixtureOfDenoisersCollator` as the collate function.'
```
dblalock, line 75: same comment as above

examples/ul2/src/mod_print_callback.py
```diff
@@ -0,0 +1,214 @@
...
+            print(
+                f'\n\n{self.__class__.__name__} printout...\n{example_string}\n',
+                file=sys.stdout,
+                flush=True)
+
+        except AssertionError:
+            if self._raise_on_error:
+                raise
+            print(
+                f'\n\n{self.__class__.__name__}: Failed to parse input/target/output strings. This is expected if the model is early in training. Set `raise_on_failure=True` if you want this to failure generate an error.\n',
```
dblalock, line 104: line length

examples/ul2/src/mod_print_callback.py
```diff
@@ -0,0 +1,214 @@
...
+                file=sys.stdout,
+                flush=True)
+
+        self._steps_since_last_print = 0
+
+    def _generate_input_target_output_strings(
+            self, batch: Dict[str, torch.TensorType],
+            model: PreTrainedModel) -> List[str]:
```
dblalock, line 112: seems like a prime candidate to be a function (or staticmethod) rather than method

examples/ul2/src/super_glue/data.py
```diff
@@ -0,0 +1,262 @@
...
+        download_config=download_config,
+    )
+
+    log.info(
+        f'Starting tokenization by preprocessing over {num_workers} threads!')
+    prefix_column_order = _task_prefix_column_names[task]
+    target_extractor_map = _task_target_maps[task]
+    if isinstance(target_extractor_map, dict):
+        target_extractor = lambda example: target_extractor_map[example['label']
+                                                               ]
```
dblalock, line 104: okay yapf, calm down

examples/ul2/src/super_glue/data.py
```diff
@@ -0,0 +1,262 @@
...
+
+        else:
+            encoder_dict = tokenizer(
+                text=encoder_text,
+                padding='max_length',
+                max_length=max_seq_length,
+                truncation=True,
+            )
+
+            decoder_dict = tokenizer(
```
dblalock, line 145:
```suggestion
            )
            decoder_dict = tokenizer(
```

examples/ul2/src/super_glue/data.py
```diff
@@ -0,0 +1,262 @@
...
+    if isinstance(target_extractor_map, dict):
+        target_extractor = lambda example: target_extractor_map[example['label']
+                                                               ]
+    else:
+        assert callable(target_extractor_map)
+        target_extractor = target_extractor_map
+
+    def convert_padding_token(input_ids: List,
+                              attention_mask: List,
+                              new_pad_id: int = -100):
```
dblalock, line 111: Looks like we need our constant to be public

```suggestion
                              new_pad_id: int = HF_IGNORE_TOKEN):
```

examples/ul2/src/super_glue/data.py
```diff
@@ -0,0 +1,262 @@
...
+        pin_memory=cfg.pin_memory,
+        prefetch_factor=cfg.prefetch_factor,
+        persistent_workers=cfg.persistent_workers,
+        timeout=cfg.timeout,
+    )
+
+
+def build_super_glue_task_dataloader(cfg: Mapping[str, Any],
+                                     device_batch_size: int,
+                                     mode: Optional[str] = None):
```
dblalock, line 201: ideally the public functions should be together at the top of the file

examples/ul2/src/super_glue/data.py
```diff
@@ -0,0 +1,262 @@
...
+            }
+
+    assert isinstance(dataset, datasets.Dataset)
+    columns_to_remove = list(dataset[0].keys())
+    dataset = dataset.map(
+        tokenize_function,
+        batched=False,
+        num_proc=None if num_workers == 0 else num_workers,
+        # batch_size=3, #1000,
+        remove_columns=columns_to_remove,
```
dblalock, line 173: just flagging this so we don't miss it
```suggestion
        num_proc=None if num_workers == 0 else num_workers,
        remove_columns=columns_to_remove,
```

examples/ul2/src/super_glue/data.py
```diff
@@ -0,0 +1,262 @@
...
+        persistent_workers=cfg.persistent_workers,
+        timeout=cfg.timeout,
+    )
+
+
+def build_super_glue_task_dataloader(cfg: Mapping[str, Any],
+                                     device_batch_size: int,
+                                     mode: Optional[str] = None):
+
+    assert cfg.name == 'super_glue', f'Tried to build super_glue dataloader with cfg.name={cfg.name}'
```
dblalock, line 203: assert vs Exceptions

examples/ul2/src/super_glue/data.py
```diff
@@ -0,0 +1,262 @@
...
+            dataset = datasets.interleave_datasets(
+                super_glue_datasets,
+                probabilities=probs,
+                seed=42,
+                stopping_strategy='all_exhausted')
+            return _build_dataloader(dataset, cfg, device_batch_size)
+
+        # Create a list of evaluators, with one evaluator per task
+        else:
+            assert mode == 'eval'
```
dblalock, line 249: assert vs Exceptions

examples/ul2/src/super_glue/data.py
```diff
@@ -0,0 +1,262 @@
...
+        return _build_dataloader(dataset, cfg, device_batch_size)
+
+    # Task mixture
+    elif isinstance(cfg.dataset.task, (list, tuple, ListConfig)):
+        tasks = [task.lower() for task in cfg.dataset.task]
+
+        if mode not in ['train', 'eval']:
+            raise ValueError(
+                'When using multiple tasks, argument `mode` must be set to either `train` or `eval`.'
+            )
```
dblalock, line 223: line length; and ideally we should print out the `mode` we received to help the user debug

examples/ul2/src/super_glue/data.py
```diff
@@ -0,0 +1,262 @@
...
+
+        super_glue_datasets = [
+            create_super_glue_dataset(task,
+                                      cfg.dataset.tokenizer_name,
+                                      cfg.dataset.split,
+                                      cfg.dataset.max_seq_length,
+                                      cfg.dataset.decoder_only_format,
+                                      extra_prefix=cfg.dataset.get(
+                                          'extra_prefix', None))
```
dblalock, line 232:
```suggestion
            create_super_glue_dataset(
                task,
                cfg.dataset.tokenizer_name,
                cfg.dataset.split,
                cfg.dataset.max_seq_length,
                cfg.dataset.decoder_only_format,
                extra_prefix=cfg.dataset.get('extra_prefix'),
```
(The diff is confused; all I did is change the linebreaks/indentation and remove the `, None` in the `get()`)

examples/ul2/src/super_glue/data.py
```diff
@@ -0,0 +1,262 @@
...
+        ]
+
+        # Merge these task datasets into one dataset, sampling from each with
+        if mode == 'train':
+            d_lengths = [len(dset) for dset in super_glue_datasets]
+            probs = [d_length / sum(d_lengths) for d_length in d_lengths]
+            dataset = datasets.interleave_datasets(
+                super_glue_datasets,
+                probabilities=probs,
+                seed=42,
```
dblalock, line 243: option to change the seed?

examples/ul2/src/super_glue/metrics.py
```diff
@@ -0,0 +1,46 @@
...
+from typing import Optional
+
+import torch
+from torchmetrics import Metric
+
+
+class ExactMatch(Metric):
+    is_differentiable = False
+    higher_is_better = True
+    full_state_update = False
```
dblalock, line 13: This class should probably be in Composer (though again, okay for now)

examples/ul2/src/utils/adapt_tokenizer.py
```diff
@@ -0,0 +1,29 @@
...
+# SPDX-License-Identifier: Apache-2.0
+
+from typing import Union
+
+from transformers import PreTrainedTokenizer, PreTrainedTokenizerFast
+
+
+def adapt_tokenizer_for_denoising(tokenizer: Union[PreTrainedTokenizer,
+                                                   PreTrainedTokenizerFast],
+                                  num_sentinel_tokens: int = 100):
```
dblalock, line 11: args docstring

examples/ul2/src/utils/adapt_tokenizer.py
```diff
@@ -0,0 +1,29 @@
...
+    """
+    # Add sentinel tokens (e.g., <extra_id_0>, <extra_id_1>, and so on). Has no effect if these are already in the vocab.
+    assert 0 <= num_sentinel_tokens <= 1000
+    sentinels_to_add = [f'<extra_id_{i}>' for i in range(num_sentinel_tokens)]
+    tokenizer.add_tokens(sentinels_to_add, special_tokens=True)
+
+    # If the padding token has not been set, add <pad> and use it
+    if tokenizer.pad_token is None:
+        tokenizer.add_tokens('<pad>', special_tokens=True)
+        tokenizer.pad_token = '<pad>'
```
dblalock, line 28: is `'<pad>'` universal enough that we can safely hardcode it? (I don't know the answer)

examples/ul2/src/utils/hf_prefixlm_converter.py
```diff
@@ -0,0 +1,769 @@
+# Copyright 2022 MosaicML Examples authors
+# SPDX-License-Identifier: Apache-2.0
+
+# Copyright 2022 MosaicML Composer authors
+# SPDX-License-Identifier: Apache-2.0
```
dblalock, line 5: double header

examples/ul2/src/utils/hf_prefixlm_converter.py
```diff
@@ -0,0 +1,769 @@
...
+
+"""Converts Huggingface Causal LM to Prefix LM.
+
+Conversion does lightweight surgery on a HuggingFace
+Causal LM to convert it to a Prefix LM.
```
dblalock, line 10: Can we move this to composer.utils or something? If it's a general-purpose utility function that we don't expect users to modify (is it?), ideally we keep it out of Examples.

examples/ul2/src/utils/hf_prefixlm_converter.py
```diff
@@ -0,0 +1,769 @@
...
+        alibi = alibi.expand(batch_size, -1, -1,
+                             -1).reshape(-1, query_length, key_length)
+        return alibi.to(dtype)
+
+    # Modified from transformers.models.bloom.modeling_bloom.BloomModel.forward
+    def forward(
+        self: BloomModel,
+        input_ids: Optional[torch.LongTensor] = None,
+        past_key_values: Optional[Tuple[Tuple[torch.Tensor, torch.Tensor],
+                                        ...]] = None,
```
dblalock, line 284: can we decompose this into a couple types with informative names? This is a looking little unreasonable + getting autoformatted poorly.

Thinking something like:
```python
KeyValueT = Tuple[torch.Tensor, torch.Tensor]
```
```python
past_key_values: Optional[Tuple[KeyValueT, ...]] = None,
```

examples/ul2/src/utils/hf_prefixlm_converter.py
```diff
@@ -0,0 +1,769 @@
...
+                          dtype=torch.int32).view(1, -1)
+        diffs = qa - ka + key_length - query_length
+        diffs = -diffs.abs()
+        alibi = slopes.view(1, num_heads, 1, 1) * diffs.view(
+            1, 1, query_length, key_length)
+        alibi = alibi.expand(batch_size, -1, -1,
+                             -1).reshape(-1, query_length, key_length)
+        return alibi.to(dtype)
+
+    # Modified from transformers.models.bloom.modeling_bloom.BloomModel.forward
```
dblalock, line 279: Would be nice to clarify that only one small portion is actually modified and link to the original source code

examples/ul2/src/utils/hf_prefixlm_converter.py
```diff
@@ -0,0 +1,769 @@
...
+    #  - new `forward` to wrap original `forward`
+    #  - new `generate` to wrap original `generate`
+    setattr(model, '_original_forward', getattr(model, 'forward'))
+    setattr(model, '_original_generate', getattr(model, 'generate'))
+
+    # Modified from transformers.models.bloom.modeling_opt.OPTDecoder._prepare_decoder_attn_mask
+    model.model.decoder.bidirectional_mask = None
+
+    def _prepare_decoder_attention_mask(self, attention_mask, input_shape,
+                                        inputs_embeds, past_key_values_length):
```
dblalock, line 590: method -> function?

dblalock (commented):
Added a bunch of comments. The macro thing is that it would be great to break this down into smaller PRs and upstream the reusable components to Composer, rather than having them in Examples. Thinking prefixLM conversion utilities and inverse sqrt scheduler in particular.

On a micro level, I commented on a *lot* of stuff, largely because I'm trying to go through and assemble contributing guidelines / coding standards and this is a great forcing function. (Relatedly, sorry if I drop into pedagogical mode in a few places.)

examples/ul2/src/utils/hf_prefixlm_converter.py
alextrott16: Sure. What is the preferred way to link it? A URL to a github repo?

examples/ul2/main.py
```diff
@@ -0,0 +1,211 @@
...
+
+def main(cfg):
+    print('Training using config: ')
+    print(om.to_yaml(cfg))
+    reproducibility.seed_all(cfg.seed)
+
+    # Read FSDP Config as a dict
+    fsdp_config = cfg.get('fsdp_config', None)
+    fsdp_config = om.to_container(fsdp_config,
+                                  resolve=True) if fsdp_config else None
```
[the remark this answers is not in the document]
alextrott16: I think we want to leave this as-is because we do want to change the `fsdp_config` data structure to a dict.

examples/ul2/src/mod_print_callback.py
alextrott16: Noted on the exceptions versus asserts. I will change that.

This callback depends on the HuggingFace `generate` method that comes with `PreTrainedModel`. It's not an interface other classes are going to have, but I will be happy to loosen the check if/when that changes.

I do think it makes sense to keep these checks here (as opposed to where the model is used) to surface any errors as early as possible. It would be a huge pain if this only threw an error after training had already made progress.

examples/ul2/src/super_glue/data.py
alextrott16: I think I can remove that, since it should be handled by the reproducibility manager.

examples/ul2/src/utils/adapt_tokenizer.py
alextrott16: This only has an effect if the tokenizer didn't set a pad token for some reason. But there are some style differences here and there. For instance, some prefer `[PAD]`. It's always the tokenizer itself that adds the padding, though, meaning nothing would break if you chose a different convention or even a non-sense string.
So as long as the token is clear to the user (which I argue `<pad>` is), I think it should be fine to make the choice to use `<pad>` if the tokenizer doesn't have a pad token set yet.

examples/ul2/src/utils/hf_prefixlm_converter.py
alextrott16: I think that should be the plan eventually, but I don't think it's mature enough for that yet.

examples/ul2/src/utils/hf_prefixlm_converter.py
alextrott16: This overwrites a method of the same name in the existing model code, so we're kinda stuck with it being a method, I think.

examples/ul2/main.py
```diff
@@ -0,0 +1,211 @@
...
+    optimizer = build_optimizer(cfg.optimizer, model)
+
+    # Scheduler
+    scheduler = build_scheduler(cfg.scheduler)
+
+    # Loggers
+    loggers = [
+        build_logger(name, logger_cfg)
+        for name, logger_cfg in cfg.get('loggers', {}).items()
+    ]
```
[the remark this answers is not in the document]
alextrott16: yapf said no :(

alextrott16:
@dblalock I committed many of your suggestions, but a number of them got re-butchered by the automatic formatter. It seems to have some weird opinions and I will do my best to get it to see reason.

examples/ul2/src/inverse_sqrt_scheduler.py
```diff
@@ -0,0 +1,31 @@
...
+from composer.optim.scheduler import ComposerScheduler
+
+__all__ = ['InverseSquareRootScheduler']
+
+
+class InverseSquareRootScheduler(ComposerScheduler):
+    """LR scheduler based on an inverse-square-root decay.
+
+    TODO(Alex): Complete docstring
+    """
```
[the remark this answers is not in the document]
alextrott16: Noted. For now, I will leave this for a different PR.

examples/ul2/src/super_glue/metrics.py
alextrott16: I'll also leave this as a future TODO for now, but noted.

alextrott16:
Closing because this was picked up in another PR.

PR #102, opened 2023-01-31
No automatic grad accum by default in BERT

alextrott16:
Incorporates the batch sizing logic from `examples/llm/main.py` into `examples/bert/main.py` and changes the associated YAMLs to work with that change.


examples/bert/yamls/test/glue.yaml
```diff
@@ -32,12 +32,12 @@ save_finetune_checkpoint_folder: ${save_finetune_checkpoint_prefix}/${base_run_n
...
 callbacks:
   lr_monitor: {}
-  speed_monitor: {}
+  # speed_monitor: {}
 
 # Scheduler
 scheduler:
   name: linear_decay_with_warmup
-  t_warmup: 0.06dur
+  t_warmup: 0.5dur
```
abhi-mosaic, line 40: is this intentional?

examples/bert/yamls/test/main.yaml
```diff
@@ -50,12 +50,10 @@ eval_loader:
   drop_last: false
   num_workers: 8
 
-global_train_batch_size: 16
-
 # Optimization
 scheduler:
   name: linear_decay_with_warmup
-  t_warmup: 0.06dur # Warmup to the full LR for 6% of the training duration
+  t_warmup: 0.5dur # Warmup to the full LR for 6% of the training duration
```
abhi-mosaic, line 56: same here

examples/bert/yamls/test/main.yaml
```diff
@@ -67,16 +65,16 @@ optimizer:
   eps: 1.0e-08
   weight_decay: 0.0
 
-grad_clip_norm: -1.0
-
```
abhi-mosaic, line 71: is it worthwhile to add the `gradient_clipping` algorithm or is that not needed for this test?

examples/bert/tests/smoketest_config_main.yaml
```diff
@@ -63,11 +61,13 @@ optimizer:
 # Training duration and evaluation frequency
 max_duration: 8ba
 eval_interval: 8ba
-eval_subset_num_batches: 8 # For code testing, evaluate on a subset of 8 batches
+# eval_subset_num_batches: 8 # For code testing, evaluate on a subset of 8 batches
```
abhi-mosaic, line 64: Is this no longer needed? The `eval_interval` above will only run `eval` once at the end of training, but it may still take a long time if the `eval_loader` is big.

examples/bert/main.py
```diff
@@ -18,6 +18,46 @@
                                       build_optimizer, build_scheduler)
 
 
+def calculate_batch_size_info(global_batch_size, device_microbatch_size):
```
abhi-mosaic, line 21: I think it might be worthwhile to move both these functions to `common/...` and then import them, in both the `bert` and `llm` examples!

abhi-mosaic (commented):
Made a couple comments, biggest Q is whether we should import the `batch_size` functions and put them in `common`

examples/bert/tests/smoketest_config_main.yaml
alextrott16: The tests use a synthetic dataset with 16 examples, so it should be good. Runs fast locally at least.

examples/bert/yamls/test/main.yaml
alextrott16: AFAIK, we've never used it with BERT. I removed this line because it is no longer referenced in the script.

examples/bert/yamls/test/glue.yaml
alextrott16: Yeah, I made the change to prevent warnings about the warmup duration being 0 batches (which was happening with the test max duration of 10 batches).

examples/bert/yamls/test/main.yaml
alextrott16: Same as above

abhi-mosaic (approved):
LGTM! Last comment -- could you explain the reasoning for two utils files `config_utils.py` and `logging_utils.py` ? It feels like we could merge them?

alextrott16:
> LGTM! Last comment -- could you explain the reasoning for two utils files `config_utils.py` and `logging_utils.py` ? It feels like we could merge them?

No good reason. I combined them and updated all the `*/main.py` files accordingly. Will merge assuming all tests pass.

PR #105, opened 2023-02-01
add throughput warning

vchiley:


abhi-mosaic (approved):
Thanks, will update JIRA to prioritize new profiling.

PR #106, opened 2023-02-01
FSDP for HF Encoder Decoder

samhavens:
This adds support for FSDP wrapping encoder-decoder HF models, and factors out the HF FSDP code into `common/`, as it will now be shared between `llm/` and `ul2/`.

It also fixes the bug in `prepare_hf_causal_lm_model_for_fsdp` where `lm_head = hf_get_causal_hidden_layers(model)` instead of `lm_head = hf_get_lm_head(model)`

Draft as I have only done limited testing. I was hoping @alextrott16 could check that this allows UL2 with FSDP.


abhi-mosaic:
I'm thinking about the `common/hf_fsdp` placement... right now our operating plan is to put UL2 underneath `llm/`. But I think it still makes sense to do this if we also add support for enc models (future) too. That way the `bert/` example could call this `prepare_hf_...` function to enable multi-billion parameter BERTs, which is missing rn.

samhavens:
Ooh will all UL2 be under `llm/`, not just MoD-finetuning? I hadn't realized that. But yeah I could see wanting to FSDP wrap an encoder-only model (T5-11B encoder is a very big, very good encoder...)

Now that you mention it, this seems like it would also be relevant for ViT as well, so I feel even better about factoring it out.

abhi-mosaic:
This generally looks good to me but it would be nice to setup some unit tests... maybe like instantiating a HF model, applying the prepare funciton, hadning to the Composer Trainer with `fsdp_config`, and verifying that `state.model` has certain layers are wrapped / no wrapped? Curious what you think.

samhavens:
Yes, I 100% support this.

alextrott16:
I will (informally) test this by plugging it into a T5 pipeline. Stay tuned.

samhavens:
I feel like (based on Slack convo that you are OOMing with t5-11b on 8x a100 40gb) something must be wrong with this. I tried to emulate the `transformers_auto_wrap_policy` per [this](https://pytorch.org/tutorials/intermediate/FSDP_adavnced_tutorial.html#transformer-wrapping-policy), but I must have missed something.

examples/common/hf_fsdp.py
```diff
@@ -0,0 +1,220 @@
...
+    # When using the HF LM models,
+    # the weights of the self.lm_head and self.transformer.wte are tied.
+    # This tying occurs inside the `self.post_init()` function.
+    # This is a hurdle for FSDP because they need to be in the same FSDP block
+    # These lines ensures that both modules stay together in the top-most block when
+    # the model has this tying enabled (almost all do; this property defaults to True)
+    if model.config.tie_word_embeddings:
+        causal_base_model._fsdp_wrap = False  # type: ignore
+        tied_embeddings._fsdp_wrap = False  # type: ignore
+        lm_head._fsdp_wrap = False  # type: ignore
```
abhi-mosaic, line 160: If you take out all of this code, and make sure that `fsdp_config.min_params` is really really big, like `1e12`, so that it doesn't get in the way, does everything still work?

alextrott16:
Following up. This "works" for T5 on the basis that:
1) It doesn't create any new errors when wrapping a `t5-base`, which doesn't require FSDP on a 40gb card
2) It allows training a `t5-3b`, which _does_ require FSDP on a 40gb card (at least for the batch sizes I was testing)

It's certainly possible that this set of tests has blind spots, but it seems to me like a good sign.

examples/common/hf_fsdp.py
```diff
@@ -0,0 +1,220 @@
...
+    # FSDP Wrap and Activation Checkpoint every decoder block
+    model.fsdp_wrap_fn = lambda module: isinstance(module, decoder_block_type)
+    model.activation_checkpointing_fn = lambda module: isinstance(
+        module, decoder_block_type)
+
+    if encoder_block_type == decoder_block_type:
+        return
+
+    # need to wrap encoder blocks separately for ProhpetNet and Marian
+    model.fsdp_wrap_fn = lambda module: isinstance(module, encoder_block_type)
```
alextrott16, line 218: Do you not want `isinstance(module, (encoder_bloc_type, decoder_block_type))`, in these two functions?

samhavens: Yeah, I should move up the equality check and do that

examples/common/hf_fsdp.py
```diff
@@ -0,0 +1,220 @@
...
+                'follow common layer/weight naming conventions.')
+    decoder_block_type = type(decoder_block[0])  # type: ignore
+    encoder_block_type = type(encoder_block[0])  # type: ignore
+
+    if model.config.tie_word_embeddings:
+        # it is possible to train an enc/dec without tied embeddings, hence the check
+        tied_embeddings._fsdp_wrap = False  # type: ignore
+        encoder._fsdp_wrap = False  # type: ignore
+        decoder._fsdp_wrap = False  # type: ignore
+        lm_head._fsdp_wrap = False  # type: ignore
```
alextrott16, line 207: Do we need to do anything special with `model.{encoder,decoder}.embed_tokens` since they are tied to `model.shared` and `model.lm_head`?

samhavens: I don't think so. `model.{encoder,decoder}.embed_tokens` are both references to `model.shared`; that is, `model.encoder.embed_tokens == model.shared`, and changing a property of the referent should change the 2 references

But, we could check this just to be safe.

That said, in the next iteration the plan is to do `isinstance(module, (encoder_bloc_type, decoder_block_type))` as the only check, which is what the [`transformer_auto_wrap_policy`](https://github.com/pytorch/pytorch/blob/c0e15fe3f4d5622c7a687cf0d503902be9296596/torch/distributed/fsdp/wrap.py#L81-L86) does

abhi-mosaic (approved):
LGTM! Lets battle test this at bug bash

PR #107, opened 2023-02-02
Uptd test attn

vchiley:


abhi-mosaic (approved):
Tested on interactive GPU instance and it's working for me, thanks @vchiley !

PR #108, opened 2023-02-02
Refactor `llm` folder

abhi-mosaic:
Adding @alextrott16 and @bmosaicml as you two are going to be adding the most functionality in the near term, and I want to see if this new structure makes things better or worse

For UL2 code, I'm thinking something like:
* `src/huggingface/...` can accept new model files we need for PrefixLM models
* `train_ul2` can be a new folder, with its own training script, its own YAMLs, and anything else needed for UL2 specific workflows. It can be used for both scratch UL2 and continued training UL2R, or really whatever you think best @alextrott16 

For ICL code, I'm thinking something like:
* `eval_icl/composer/` or even just `eval_icl/...` can be used to hold any standalone eval_scripts, README, throughput tables, etc. 
* any importable components should go in `src/...`. But there might not be any? I think we may just need changes to `mosaic_gpt.py` and the training script `train_clm.py` ?

Let me know what you think! We should sync tomorrow so we can get this in soon and pave the way for all the good UL2 and ICL work :) 

P.S. the best way to navigate the new folder and see what it feels like is to start at this link: https://github.com/mosaicml/examples/tree/abhi/refactor_llm/examples/llm


bmosaicml (approved):
Looks good to me! When I merge my ICL code we can deprecate the Eleuther code & import

examples/llm/train_clm/train_clm.py
```diff
@@ -13,20 +13,29 @@
...
-        message='Torchmetrics v0.9 introduced a new argument class property')
-    try:
-        return COMPOSER_MODEL_REGISTRY[cfg.name](cfg)
-    except:
+    if cfg.name == 'mosaic_gpt':
+        return ComposerMosaicGPT(cfg)
+    elif cfg.name == 'hf_causal_lm':
+        return ComposerHFCausalLM(cfg)
+    else:
         raise ValueError(f'Not sure how to build model with name={cfg.name}')
```
alextrott16, line 26: I know this isn't part of the PR, but if you feel like tacking on a minor TODO, we should specify the valid options in these `ValueError`s.

Alternatively, feel free to ignore this comment for now and maybe lets make a separate JIRA to do this throughout `examples`.

examples/llm/train_clm/yamls/hf_causal_lm/gpt-neo-125m.yaml
```diff
@@ -85,11 +85,10 @@ precision: amp_bf16
 # FSDP
```
alextrott16, line 85: Perhaps outside the scope of this PR, but is it actually recommended to use FSDP on such lightweight models?

If no, then I might recommend that, where appropriate, we comment out the `fsdp_config` blocks and leave a comment saying something along the lines of "you can comment/uncomment this block to turn FSDP off/on; it is typically not required for the workload in this config".

alextrott16 (approved):
Overall, this looks good to me and the structure makes sense. I wouldn't be surprised if we end up doing another refactor as new code begins to land inside `llm/`, but I think that's mostly fine.

A few notes:
- I left a couple comments, both somewhat out of scope for this PR. Feel free to ignore them if you'd prefer to revisit them in a different PR.
- I'm assuming that a hefty refactor of `llm/eval_icl` will take place when we upgrade to `composer 0.12.1`. I think that will be a good thing to do and will hopefully cut down on the need for `llm/src/model_loading.py` and `llm/src/tokenizer.py`. Actually, I wonder if those should go into `llm/eval_icl` since it seems like that's where they're used. **Please consider whether that makes sense to you too before merging.**
- The more I think about it, the more I realize that `composer 0.12.1` will likely land a number of changes that will warrant another (hopefully small) refactor and/or simplification of the code. For example, it will allow us to use `HuggingFaceModel` instead of having to wrap it in our custom `ComposerHFCausalLM` in order to handle label shifting.

These notes are just something to consider for future work in the direction of this PR. I'm happy to approve the existing changes as is, if you feel it's better to address these items down the road.

abhi-mosaic:
> The more I think about it, the more I realize that composer 0.12.1 will likely land a number of changes that will warrant another (hopefully small) refactor and/or simplification of the code. For example, it will allow us to use HuggingFaceModel instead of having to wrap it in our custom ComposerHFCausalLM in order to handle label shifting.

Absolutely. I want to demonstrate using the `HuggingFaceModel` as much as possible so that our customers do too.

> Actually, I wonder if those should go into llm/eval_icl since it seems like that's where they're used. Please consider whether that makes sense to you too before merging.

I'll think about it more... some of the model loading code is sort of like "make it easy to fetch an S3 checkpoint and load into a model class". Those helpers might be come in handy later? But I don't know for sure... 

I agree on the `tokenizer.py` stuff. I want to remove as many registries / abstractions as much as possible, and since we don't plan on using anything but HF tokenizers for now, I would prefer to remove the file if possible.

alextrott16:
> I agree on the tokenizer.py stuff. I want to remove as many registries / abstractions as much as possible, and since we don't plan on using anything but HF tokenizers for now, I would prefer to remove the file if possible.

FWIW, I thought about making the same recommendation wrt removing `tokenizer.py`, but figured better not to mention that for now to avoid more scope creep.

examples/llm/train_clm/yamls/hf_causal_lm/gpt-neo-125m.yaml
abhi-mosaic: I think.. its probably not necessary, but it doesn't hurt... if someone is trying this on some low resource cards at home, maybe it still help them out.

I will make a JIRA for this but not address in this PR

PR #110, opened 2023-02-02
Restore Docker image workflow

abhi-mosaic:
This Github workflow will push `mosaicml/examples:[bert,llm]` images with the example requirements pre-installed.

### DIFF
--- a/.github/workflows/build_docker.yaml
+++ b/.github/workflows/build_docker.yaml
@@ -1,0 +1,52 @@
+name: Docker build
+on:
+  push:
+    branches:
+    - main
+  workflow_dispatch: {}
+jobs:
+  docker-build:
+    runs-on: ubuntu-latest
+    if: github.repository_owner == 'mosaicml'
+    strategy:
+      fail-fast: false
+      matrix:
+        example:
+        - bert
+        - llm
+    steps:
+    - name: Checkout
+      uses: actions/checkout@v3
+
+    - name: Setup QEMU
+      uses: docker/setup-qemu-action@v2
+
+    - name: Setup Docker Buildx
+      uses: docker/setup-buildx-action@v2
+
+    - name: Login to DockerHub
+      uses: docker/login-action@v2
+      with:
+        username: ${{ secrets.DOCKER_HUB_USERNAME }}
+        password: ${{ secrets.DOCKER_HUB_PASSWORD }}
+
+    - name: Calculate Docker Image Variables
+      run: |
+        set -euxo pipefail
+
+        ###################
+        # Calculate the tag
+        ###################
+        GIT_SHA=$(echo ${{ github.sha }} | cut -c1-7)
+        echo "IMAGE_TAG=${GIT_SHA}" >> ${GITHUB_ENV}
+
+    - name: Build and Push the Docker Image
+      uses: docker/build-push-action@v3
+      with:
+        context: .
+        tags: mosaicml/examples:${{ matrix.example }}-latest,
+          mosaicml/examples:${{ matrix.example }}-${{ env.IMAGE_TAG }}
+        push: ${{ contains(github.ref, 'refs/heads/main') }}
+        cache-from: type=registry,ref=mosaicml/examples:${{ matrix.example }}-buildcache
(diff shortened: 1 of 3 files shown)


Dockerfile
```diff
@@ -4,7 +4,9 @@
...
 
 ARG EXAMPLE
-COPY ${EXAMPLE}/requirements.txt requirements.txt
+COPY requirements.txt requirements.txt
+COPY examples/${EXAMPLE}/requirements.txt ${EXAMPLE}-requirements.txt
 
 RUN pip install --no-cache-dir -r requirements.txt && \
-    rm requirements.txt
+    pip install --no-cache-dir -r ${EXAMPLE}-requirements.txt && \
+    rm *.txt
```
dblalock, line 12:
```suggestion
    rm requirements.txt ${EXAMPLE}-requirements.txt
```
Forward compatibility in case we (or a user's fork) adds any other txt files

.github/workflows/build_docker.yaml
```diff
@@ -0,0 +1,52 @@
...
+  pull_request:
+    branches:
+    - main
+  workflow_dispatch: {}
+jobs:
+  docker-build:
+    runs-on: ubuntu-latest
+    if: github.repository_owner == 'mosaicml'
+    strategy:
+      fail-fast: false
```
dblalock, line 12: Is there a reason we're overriding the default behavior here? ([fail-fast docs](https://docs.github.com/en/actions/using-workflows/workflow-syntax-for-github-actions#jobsjob_idstrategyfail-fast))

dblalock (approved):
Basically lgtm. fail-fast thing is just an optimization so probably fine to cargo-cult it from wherever you copied this from

.github/workflows/build_docker.yaml
abhi-mosaic: I am not sure... this is from @bandish-shah , will let him comment

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,50 @@
+name: Docker
+on:
+  pull_request: {}
```
abhi-mosaic: Btw, I know this was originally on `push` to `main`, but it seems to be so fast due to cacheing, that I wonder if it's ok to just run this on PR? It would be helpful in testing I think.

.github/workflows/docker.yaml
```diff
@@ -0,0 +1,52 @@
+name: Docker
+on:
+  pull_request:
+    branches:
+    - main
```
abhi-mosaic, line 5: Btw, I know this was originally on `push` , but since its running really fast, is it ok to just have this run on every `pull_request` to main? @dblalock

dblalock: will that result in every PR overwriting the image on dockerhub? slash, could someone just open some random PR that corrupts the image and have it distributed to everyone who docker pulls it?

.github/workflows/build_docker.yaml
bandish-shah: That wasn't intentional probably just a copy/paste artifact from whatever flow I originally based this on.  You can enable it if you want, doesn't really matter right now since the build is so quick and you're only building 2 images.

.github/workflows/docker.yaml
abhi-mosaic: After discussing from Bandish, my understanding of what will happen now is:
* inside PR, every commit will **build** but **not push** a new image. So you can check with Github Action CI if the image builds or breaks. There is not a direct way to try running on the image since it is not pushed
* On PR approval and merge,  it will **rebuild** and **push** to `mosaicml/examples:[llm]-[main-commithash]` and `mosaicml/exmaples:[llm]-[latest]`. 

So we will not overwrite any tags that have a hash. And we will only overwrite the `-latest` ones.

.github/workflows/build_docker.yaml
abhi-mosaic: I will revert to default as we are doing this on PRs

abhi-mosaic:
Going to merge this in 5 so I can see how the new image does, will open a second PR if anything bad happens.

PR #111, opened 2023-02-03
Seed_all in resnet example

Landanjs:


coryMosaicML (approved):
LGTM! Confirmed this fixes things

karan6181:
@Landanjs wondering the reason behind pinning the old streaming version `0.2.1`? Is it because of some hang that you have observed earlier ?

bandish-shah (changes requested):
Would it be possible to use the seed workaround and not downgrade streaming?

dakinggg:
Yeah, landan just pushed that change, i think at the same time you commented.

bandish-shah (approved):
Beat me to it!  Thanks!

examples/resnet/main.py
```diff
@@ -33,6 +33,7 @@ def build_logger(name: str, kwargs: Dict):
 
 
 def main(config):
+    reproducibility.seed_all(cfg.seed)
```
dakinggg, line 36:
```suggestion
    reproducibility.seed_all(config.seed)
```

examples/resnet/requirements.txt
```diff
@@ -1,2 +1,2 @@
-mosaicml[streaming,wandb]>=0.11.0,<0.13
+mosaicml[wandb]>=0.11.0,<0.13
```
dakinggg, line 1:
```suggestion
mosaicml[streaming,wandb]>=0.11.0,<0.13
```

PR #114, opened 2023-02-04
add how to extend section to readme

dakinggg:
I decided it didn't really make sense to add an example of how to extend things to every example's individual readme, because they would all be basically the same. I think the basic steps are fairly clear, so I didn't go into great detail, but definitely open to ideas for improving it, or if someone had a different vision for how this should look.


dblalock (approved):
Ended up adding some stuff but just gonna approve since I don't think we need another round trip for this.

PR #119, opened 2023-02-06
Speed monitor supports no cuda

alextrott16:
This adds a simple fix to make `common/speed_monitor_w_mfu.py` work if there is no CUDA device.

### DIFF
--- a/examples/common/speed_monitor_w_mfu.py
+++ b/examples/common/speed_monitor_w_mfu.py
@@ -126,6 +126,9 @@ def get_gpu_flops_available(state: State):
             f'MFU cannot be calculated and reported. gpu_flops_available can be manually' +\
             f'overridden by setting gpu_flops_available in SpeedMonitorMFU()'
         )
+        # Setting to 0 will disable MFU computation and prevent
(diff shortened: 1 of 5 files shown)


examples/common/speed_monitor_w_mfu.py
```diff
@@ -172,8 +177,8 @@ def before_dataloader(self, state: State, logger: Logger) -> None:
         self.batch_start_num_samples = int(state.timestamp.sample)
 
         # Get available GPU FLOPS
-        if not self.gpu_flops_available:
-            self.gpu_flops_available = get_gpu_flops_available(state)
+        if self.gpu_flops_available is None:
+            self.gpu_flops_available = get_gpu_flops_available(state) or 0
```
abhi-mosaic, line 181: I'm guessing that without this, the value stays `None` and keeps trying to be reset on every batch, is that right?

I think it might be a bit better to move the `or 0` to L130 so that the 0 logic is only there rather in both places?

abhi-mosaic (approved):
LGTM! Thanks @alextrott16

PR #121, opened 2023-02-06
Patch MosaicBERT for 0.12.1

alextrott16:
Bumps `bert` code to composer 0.12.1 and addresses an issue stemming from MosaicBERT only supporting `use_return_dict = False` and requiring that `labels` are passed into the forward method. This PR adds MosaicBERT support for both of those use cases, as now required for using `HuggingFaceModel` in composer 0.12.1.

Note: I can successfully reproduce past convergence runs that used hardcoded `use_return_dict = False`, so that's good.

### DIFF
--- a/examples/bert/src/bert_layers.py
+++ b/examples/bert/src/bert_layers.py
@@ -815,11 +815,10 @@ class BertForMaskedLM(BertPreTrainedModel):
         if (input_ids is not None) == (inputs_embeds is not None):
             raise ValueError('Must specify either input_ids or input_embeds!')
 
-        masked_lm_labels = labels
-        if masked_lm_labels is None:
+        if labels is None:
             masked_tokens_mask = None
         else:
-            masked_tokens_mask = masked_lm_labels > 0
+            masked_tokens_mask = labels > 0
 
         return_dict = return_dict if return_dict is not None else self.config.use_return_dict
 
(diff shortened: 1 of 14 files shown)


examples/bert/src/bert_layers.py
```diff
@@ -815,13 +815,12 @@ def forward(
...
-            raise ValueError('Mosaic BertForMaskedLM requires a labels tensor.')
-
-        return_dict = return_dict if return_dict is not None else self.config.return_dict
+        if labels is None:
+            masked_tokens_mask = None
+        else:
+            masked_tokens_mask = labels > 0
 
-        masked_tokens_mask = masked_lm_labels > 0
+        return_dict = return_dict if return_dict is not None else self.config.use_return_dict
```
abhi-mosaic, line 823: does this work? I've been seeing this syntax used by others and I feel like it helps reduce line length, haha
```suggestion
        return_dict = return_dict or self.config.use_return_dict
```

dakinggg: ^I don't think this works because 
```
In [1]: False or True
Out[1]: True
```

alextrott16: Oooh, good catch @dakinggg. I was also under the impression that that `a or b` was a special syntax for `a if a is not None else b`. Basically, you have to keep the long version because `return_dict` has a type signature of `Optional[bool]`. Today I learned...

dakinggg: haha I try to avoid Falseyness in python as much as possible. So many subtle bugs

dakinggg (approved):
LGTM, one comment that you are free to take or leave. In the HF code, they just have one
```
if not return_dict:
    optionally put the loss in the tuple depending on whether loss is None
```

In your code you have that conditional twice, in each larger conditional block of `if labels is None`. Might be worth trying to stick with the HF code structure as much as possible? But as I said, fine either way.

PR #122, opened 2023-02-07
Fix ci

dakinggg:
A few things to note:
- added mmcv-full dependency for deeplab (cc @Landanjs @A-Jacobson should I pin the version of that and mmseg?)
- use our docker images so that I can get the vision image for deeplab tests
- SAM appears to have a bug when run without grad scaling, skipping the test for now (@Landanjs is fixing in composer)
- streaming seems to have a bug. haven't totally root caused, but increasing the dataset size from 16 -> 128 resolves it for now. root cause is still being investigated.

### DIFF
--- a/.github/workflows/test.yaml
+++ b/.github/workflows/test.yaml
@@ -22,14 +22,19 @@ jobs:
         include:
         - name: bert-pytest
           container: mosaicml/pytorch:1.13.1_cpu-python3.10-ubuntu20.04
+          subdir: bert
         - name: cifar-pytest
           container: mosaicml/pytorch_vision:1.13.1_cpu-python3.10-ubuntu20.04
+          subdir: cifar
         - name: deeplab-pytest
           container: mosaicml/pytorch_vision:1.13.1_cpu-python3.10-ubuntu20.04
+          subdir: deeplab
         - name: llm-pytest
           container: mosaicml/pytorch:1.13.1_cpu-python3.10-ubuntu20.04
+          subdir: llm
         - name: resnet-pytest
           container: mosaicml/pytorch_vision:1.13.1_cpu-python3.10-ubuntu20.04
+          subdir: resnet
     name: ${{ matrix.name }}
     container: ${{ matrix.container }}
     steps:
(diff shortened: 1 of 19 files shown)


.github/workflows/test.yaml
```diff
@@ -19,34 +19,32 @@ jobs:
...
-          ${{ runner.os }}-
     - name: Setup
       run: |
         set -ex
-        python -m pip install --upgrade pip wheel
+        export PATH=/composer-python:$PATH
+        python -m pip install --upgrade 'pip<23' wheel
     - name: Test
       run: |
+        export PATH=/composer-python:$PATH
```
dblalock, line 49: is there a clean way to avoid adding this to the path? I doubt users will modify their PATH, so I'm worried about the CI not reflecting what users experience if they try to install and use the repo

dblalock (approved):
Basically LGTM. The one thing I worry about is having the CI not reflect the user's environment, but maybe there's no good way around this? Or we just punt this for now?

I also like the idea of testing that our install requirements work outside of our recommended docker images...but we also need to test the _inside_ our recommended docker images, so maybe we just ideally have a few more entries in the matrix. Sounds like something we could address outside of this PR though.

.github/workflows/test.yaml
dakinggg: So the path adding is only needed because composer images do weird stuff with python. I think we're going to fix this soon and then this command goes away. That being said, the actual lint/test is still running inside a venv, so that should be basically what a user would experience.

dakinggg:
Yeah, we can add more tests later that don't user our docker images (made a JIRA). Installing the vision stuff is a bit of a challenge, which is the reason I switched to using our images.

abhi-mosaic:
> but increasing the dataset size from 16 -> 128 resolves it for now. root cause is still being investigated.

This is probably due to `num_canonical_nodes` defaulting to 128. If there are less than that many samples, then it's not possible to create 1 batch. Still not 100% sure if this is intended behavior by `StreamingDataset`.

dakinggg:
> > but increasing the dataset size from 16 -> 128 resolves it for now. root cause is still being investigated.
> 
> This is probably due to `num_canonical_nodes` defaulting to 128. If there are less than that many samples, then it's not possible to create 1 batch. Still not 100% sure if this is intended behavior by `StreamingDataset`.

I didn't see any where in the bert code that it was defaulting to 128 though. Where is that happening?

dakinggg:
Ah I think I was looking at an old version of the code but CI was running on a newer version. Sounds good.

PR #123, opened 2023-02-07
Upgrade to `mosaicml-streaming==0.2.3`

abhi-mosaic:
This PR updates the BERT and LLM examles to use `mosaicml-streaming==0.2.3`.

We also fix an outstanding TODO to set the default value for `num_canonical_nodes` to a high number, as this will determine the shuffling randomness. The rule is "the samples of the first batch come from the first `num_canonical_nodes` shards", and then samples are approximately distributed like that for the rest of the epoch.

I've verified that on 1 node on OCI, the additional load time from `num_canonical_nodes=1 -> num_canonical_nodes=128` for The Pile dataset is <15s. I think this is a reasonable overhead given how long these LLM training runs usually are.

No breaking changes are expected from the minor release, and tests for dataset creation + training are passing. 

Screenshot 2023-02-06 at 7 52 30 PM

### DIFF
--- a/examples/bert/requirements.txt
+++ b/examples/bert/requirements.txt
@@ -2,7 +2,7 @@ datasets==2.7.1
 einops==0.5.0
 torch==1.13.1
 mosaicml==0.12.1
-mosaicml-streaming==0.2.2
+mosaicml-streaming==0.2.3
 omegaconf==2.2.3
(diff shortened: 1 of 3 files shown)


dblalock (approved):
LGTM

PR #124, opened 2023-02-07
Upgrade flash-attn==0.2.8

abhi-mosaic:
Needed for throughput testing. Seeing significant memory reduction vs. `flash-attn==0.2.2`


dblalock (approved):
LGTM

PR #125, opened 2023-02-07
updt attn warnings

vchiley:
While `attn_impl: triton` can be faster than `attn_impl: flash` it uses more memory. When training larger models this can trigger alloc retries which hurts performance. If encountered, we recommend using `attn_impl: flash`.

Adding a warning to note this fact.


abhi-mosaic (approved):
LGTM, thank you!

PR #126, opened 2023-02-07
Fix grad_accum reporting logic

abhi-mosaic:
When we use a `device_train_microbatch_size` that is not a divisor of `device_train_batch_size`, there will be 1 additional smaller microbatch.

The reporting logic (not used in Trainer anywhere, just configs+logs) does not account for this.

The diff in the `parse_logs.py` looks big but it's just the linter reformatting the lines.


alextrott16 (approved):
LGTM

PR #127, opened 2023-02-07
Updt attn

vchiley:
This PR updates attention in the examples/llm subdir primarily focusing on triton attn

## Changes
- Given we install flash_attn, I import the triton kernel from the flash_attn instead of having a copy of the kernel.
  - If / when we decide to make changes to the triton flash_attn kernel itself, then we can start maintaining an internal version.
- updt triton dev build from `triton==2.0.0.dev20221103` to `triton==2.0.0.dev20221202`
  - this is the version used by the flash_attn repo (see [here](https://github.com/HazyResearch/flash-attention/blob/33e0860c9c5667fded5af674882e731909096a7f/training/Dockerfile#L81))
- we do not do apply key_padding_mask when it is not needed.
  - In right padded causal attn, the token interaction with future tokens is masked out by the causal structure to `-inf`. For causal attention, it is only necessary to apply key_padding_mask if there are valid tokens _**after**_ the first invalid token.
  - I also changed the `fill_value` used by the `key_padding_mask` to mimic what bert code does
    - potentially more numerically stable
 
### Result of removing bias when not needed:
When using `group_method: concat` training curve for triton vs flash is exactly the same
Screenshot 2023-02-07 at 10 01 46 AM
Zooming in, you can see there are slight numerical diffs, but its tiny
Screenshot 2023-02-07 at 10 04 44 AM

When using `group_method: truncate` training curve for triton vs flash is exactly the same
Screenshot 2023-02-07 at 10 06 23 AM
Zooming in, you can see there are slight numerical diffs, but its tiny
Screenshot 2023-02-07 at 10 06 53 AM

#### `group_method: truncate` vs using `group_method: concat`
produces diff results which can be best seen in the eval curves (this is expected)
Screenshot 2023-02-07 at 10 08 09 AM

#### Alibi unaffected
Alibi still works / does well!
##### `group_method: truncate`
Screenshot 2023-02-07 at 11 43 34 AM

##### `group_method: concat`
Screenshot 2023-02-07 at 11 42 54 AM


examples/llm/src/flash_attention.py
```diff
@@ -25,16 +25,15 @@ class FlashAttention(nn.Module):
     """
 
     def __init__(self, num_heads, softmax_scale=None, device=None, dtype=None):
```
abhi-mosaic, line 27: is this entire file still needed? Or can we get away with a single import from `flash-attn` with a `triton` flag?

Looking at this module here: https://github.com/HazyResearch/flash-attention/blob/780e8eeabb84fe3f41e8244f04521743b032ba35/flash_attn/modules/mha.py#L39

vchiley: Yeah I wanted to do that, but that version does not enable bias; this means we can't use alibi and have no way of handleing key_pad_mask (when we need to handle it)

abhi-mosaic (approved):
Small comment on the `triton` import. Relying on your testing + the GPU tests to check that nothing broke on the `key_padding_mask` side.

examples/llm/requirements.txt
```diff
@@ -5,7 +5,7 @@ mosaicml-streaming==0.2.3
 flash-attn==0.2.8
 mosaicml-cli>=0.2.32,<1
 # need a newer version of triton
-triton==2.0.0.dev20221103
+triton==2.0.0.dev20221202
```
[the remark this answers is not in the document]
vchiley: Is that communicated to us in anyway?
The only place this is mentioned as a req is in the [readme](https://github.com/HazyResearch/flash-attention/tree/main/training#requirements) for the `training/` subdir and its part of the dockerfile [here](https://github.com/HazyResearch/flash-attention/blob/main/training/Dockerfile#L81)

I'm not sure flash-attn communicates requirements in any way...
I'l try to install / run the tests it without the requirement.

PR #128, opened 2023-02-07
Update scripts to pre-concat text and tokens

samhavens:
Pre-concatenating based on text or tokens. Fixed the encoding issues; thanks `numpy`! There is probably something to be documented about how to serialize tensors to bytes for streaming...

Allows for streaming processing and `train_small` subset. There is a performance hit (at least on Mac) when tokenizing (~80x slower than not tokenizing)

Example commands:

* No pre-concat: `python convert_c4.py --out_root=out/folder`
* Pre-concat text: `python convert_c4.py --out_root=out/folder --splits val train_small --concat_text=5000 --bos_text="</s>"`
* Pre-concat tokens: `python convert_c4.py --out_root=out/folder --splits val train_small --concat_tokens=4096 --tokenizer=facebook/opt-125m`
* Pre-concat tokens (gpt2 works now): `python convert_c4.py --out_root=out/folder --splits val train_small --concat_tokens=4096 --tokenizer=gpt2 --bos_text="<|endoftext|>"`

Note that `--splits` is variadic so we can't use `=` with it

To use pre-concatenated text is easy,

```python
from streaming.base import StreamingDataset

ds = StreamingDataset(local="out/folder", split="val")

for sample in ds:
    concatenated_text = sample['text']
```

For pre-concatenated tokens:

```python
import torch
from streaming.base import StreamingDataset
from transformers import AutoTokenizer

# this has to be the same tokenizer they used to create the data
# do we want to enforce this?
tokenizer = AutoTokenizer.from_pretrained("your/tokenizer)

ds = StreamingDataset(local="mds-data-folder", split="val")

for sample in ds:
  # note, you need to copy the numpy array because the original is non-writeable
  # and torch does not support non-writeable tensors, so you get a scary warning and
  # if you do try to write to the tensor you get undefined behavior
  tokens = torch.from_numpy(np.frombuffer(ds[0]['tokens'], dtype=np.int64).copy())
  text = tokenizer.decode(tokens)
```


samhavens:
Converting this to use streaming is rough, the signature of the map function changes, and something about the mapping / processing I do breaks the dataset iterator...

examples/common/convert_c4.py
```diff
@@ -111,23 +365,39 @@ def main(args: Namespace) -> None:
...
+
+        # When concatenating text I notice a ~6x slowdown the first time I ran this:
+        #   no concat iterations/s = ~6800 it/s
+        #   with length 10_000 char, that's ~5x longer sequences,
+        #   so we'd expect ~1350 it/s, but observe ~220 it/sec
+        # However, subsequent runs were 5x faster... ¯\_(ツ)_/¯
+        #
+        # Concatenating tokens is, at least on a Mac, much worse
+        #   I got 85 it/s, despite their being half as many sequences
+        #   which is an 80x slowdown
```
abhi-mosaic, line 397: Maybe worth removing the comments from the file, or cleaning up a bit?

FWIW, I think concatenating text is not ~5x longer sequences, it's probably more like 8-10x longer sequences as the average seq length is quite small. What I would care most about -- on a Linux machine with 64 cores, what's the total time to process `validation` when using `NO_CONCAT` vs. `CONCAT_TEXT`? As long as both are around ~10sec I wouldn't be worried.

samhavens: I will check on a linux box with more cores and remove these comments

samhavens: `CONCAT_TEXT` took 9s on a 64 core machine

examples/common/convert_c4.py
```diff
@@ -4,105 +4,359 @@
...
-    args.add_argument('--splits',
-                      nargs='+',
-                      default=['train', 'train_small', 'val'])
+    parser = ArgumentParser(description='Convert C4 into MDS format, optionally concatenating')
+    parser.add_argument('--out_root',
+                        type=str,
+                        required=True,
+                        action=EmptyOrNonexistentDirectory)
+    parser.add_argument('--compression', type=str, default=None)
+    parser.add_argument('--splits',
```
dakinggg, line 46: multiple splits doesn't seem to be working (mentioned in dm)

dakinggg (commented):
Haven't reviewed the code yet, just tried it out and works ok for me. confirmed training starts ok with truncate method and concat text or no concat. haven't tried concat tokens cause its not implemented yet. also haven't checked that the data is "right"

examples/common/convert_c4.py
```diff
@@ -4,105 +4,359 @@
...
+        self.batch_size = batch_size
+        self._build_loader()
+
+    def num_shards(self):
+        it = self.dataset._ex_iterable  # type: ignore
+        return len(it.kwargs['filepaths'])  # type: ignore
+
+    def __iter__(self):
+        worker_info = get_worker_info()
+        if worker_info:
```
dakinggg, line 119:
```suggestion
        if worker_info is not None:
```

examples/common/convert_c4.py
```diff
@@ -4,105 +4,359 @@
...
-def build_hf_c4_dataset(split: str) -> IterableDataset:
+        Args:
+            samples (IterableDataset): An iterable dataset that is multi-worker compatible
+
+        Yields:
+            Sample dicts.
+        """
+
+        n_samples = 0
+        buffer = {}
```
dakinggg, line 194: seems like this doesn't really need to be a dict and can just be a single string?

examples/common/convert_c4.py
```diff
@@ -4,105 +4,359 @@
...
+        while True:
+            for batch in self.loader:
```
dakinggg, line 251: so i think we probably need to parallelize this one. 10m for the val split is not ideal. thoughts?

examples/common/convert_c4.py
```diff
@@ -4,105 +4,359 @@
...
+                encoded = self.tokenizer(batch['text'], truncation=False, padding=False)
+                for iids in encoded['input_ids']:
+                    buffer['tokens'] = buffer.get('tokens', []) + iids
```
dakinggg, line 254: same comment as above, this can just a be a single list, right?

dakinggg, line 254: should there be a sep token here too?

examples/common/convert_c4.py
```diff
@@ -4,105 +4,359 @@
...
+                        concat_sample['tokens'] = buffer['tokens'][:self.
+                                                                max_length]
+                        buffer['tokens'] = buffer['tokens'][self.max_length:]
+
+                        yield {
+                            # convert to bytes to store in MDS binary format
+                            'tokens': np.asarray(concat_sample['tokens']).tobytes()
+                        }
+                        n_samples += 1
+                        if n_samples == self.expected_num_samples:
```
dakinggg, line 266: do we need to dump whatever is left in the buffer at the end?

examples/common/convert_c4.py
```diff
@@ -4,105 +4,359 @@
...
+        max_length: Optional[int] = None,
+        sep_text: Optional[str] = None,
+        tokenizer: Optional[PreTrainedTokenizerBase] = None,
+        expected_num_samples: Optional[int] = None) -> IterableDataset:
     """Collect the samples for this dataset split.
 
     Args:
         split (str): Split name.
+        mode (ConcatMode): NO_CONCAT, CONCAT_TEXT, or CONCAT_TOKENS
+        sep_text (str): if mode is CONCAT_TEXT, what string to use to separate samples
```
dakinggg, line 282: left comment above, but maybe here is a better place. should `CONCAT_TOKENS` also support `sep_text`?

examples/common/convert_c4.py
```diff
@@ -4,105 +4,359 @@
...
+        tokenizer (PreTrainedTokenizerBase): if mode is CONCAT_TOKENS, the tokenizer to use
+        expected_num_samples (int): _not yet implemented_ once we have generated this many samples, quit
```
dakinggg, line 284: looks like this is implemented?

examples/common/convert_c4.py
```diff
@@ -4,105 +4,359 @@
...
-            self.dataset = hf_datasets.load_dataset(path='c4',
-                                                    name='en',
-                                                    split=split,
-                                                    streaming=True)
+def _concat_sample_count(sample_count: int,
+                         mode: ConcatMode,
+                         max_length: Optional[int] = None) -> int:
+    emp_chars_per_sample = 2163  # measured for val split
+    est_tokens_per_samples = 540  # using est_char_per_token = 4
+    match mode:
```
dakinggg, line 309: this forces us to python 3.10, so maybe lets just use if/else?

examples/common/convert_c4.py
samhavens: I was imagining relying on the tokenizer to add special tokens, but this may be around the way GPT2's tokenizer does not do that. Is that what you are thinking?

examples/common/convert_c4.py
samhavens: good point on the version. it was finally a use case for matching with guards and I got excited

examples/common/convert_c4.py
samhavens: This should be working now, updating the example commands

examples/common/convert_c4.py
dakinggg: so you assume that the tokenizer adds special tokens including EOS? we should probably at least have a way to not require that? but i haven't thought about this as much as you, so trust what you think.

examples/common/convert_c4.py
samhavens: I was assuming it would get GC'ed but I can dump it explicitly

dakinggg: er, by "dump" i meant "yield"

examples/common/convert_c4.py
samhavens: the default behavior of huggingface tokenizers is to pass `add_special_tokens=True`... but we know that e.g. GPT2 does not add EOS or BOS. If the tokenizer you pass is like GPT2 and it doesn't add either, currently I raise an exception, so we can be sure that we won't pre-tokenize concat without separators. But to support GPT2, we'd need to do what you suggest and if they pass `--sep_text="<|endoftext|>"`, add that.

we should also let them pass `--sep_text_position`, right now I assume they want BOS

dakinggg: all good points. ok, as long as you error and not silently fail. seems like something we can fix after the bones of this are in.

dakinggg: although maybe we really need to support the gpt2 tokenizer? not sure

samhavens:
UPDATE: I added some changes to `text_data.py` that (based on very limited testing) seem to work with pre-concatenated text and tokens. No config changes are needed to use it

examples/common/text_data.py
```diff
@@ -23,8 +27,8 @@ class StreamingTextDataset(StreamingDataset):
         tokenizer_name (str): The name of the HuggingFace tokenizer to use to
             tokenize samples.
         max_seq_len (int): The max sequence length of each sample.
-        group_method (str): How to group text samples into token samples.
-            Supports 'truncate' or 'concat'.
+        group_method (str): (Deprecated, use ``concat`` option in create_c4.py) How to group text
```
dakinggg, line 30: this is now "use --concat_text or --concat_tokens" right?

examples/common/text_data.py
```diff
@@ -108,15 +121,52 @@ def _tokenize(self, text_sample):
...
+            tokens = self.tokenizer(text_sample['text'],
+                                    truncation=truncation,
+                                    padding=padding,
+                                    max_length=max_length,
+                                    add_special_tokens=True)
+            if self.group_method == 'concat':
+                # True is the default value so this will not change any behavior
+                self.add_special_tokens = True
+            else:
+                # if the 2nd token is BOS or 2nd-to-last-is EOS, the special tokens
```
dakinggg, line 137: why 2nd and 2nd to last instead of first and last? im probably just forgetting something here...

samhavens: Because if the tokenizer adds a BOS token AND we have used that BOS token as our `sep_text` then we will see BOS in position 1, similar for EOS in position -2

samhavens: I had not considered the case where sep_text is not EOS or BOS. Do you think it's okay to assume it is one of those?

samhavens: OK, switched sep_text to bos_text and eos_text

examples/common/text_data.py
```diff
@@ -64,6 +68,13 @@ def __init__(self,
                  batch_size: Optional[int] = None):
 
         # Validation
+        warning_msg = textwrap.dedent("""\
```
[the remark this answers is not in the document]
samhavens: done

examples/common/convert_c4.py
samhavens: I changed sep_text to bos_text and eos_text and didn't make it exclusive with tokenization, so now we can handle GPT2 if the user gives us `--bos_text=<|endoftext|>`

examples/common/convert_c4.py
samhavens: Reasonable question. Unless you have strong feelings otherwise, I think not.

We _do_ end up with sequences that start without a BOS token... but those are the sequences that start not at the beginning of a true sequence.

examples/common/convert_c4.py
samhavens: I don't think so. We've reached the "end"

dakinggg: sorry this was the most confusing comment. what I _actually_ meant was, is there a case where we don't return at this statement, but instead complete the for loop, and then we should yield the last partial sample?

samhavens: I don't think we ever want to yield the partial sample, because the user specified how long each sample should be, and it won't be that long (well, 1 in max_length times I guess it will be)

dakinggg: got it, sounds good. sorry for the back and forth on this one :)

examples/common/text_data.py
```diff
@@ -64,6 +68,14 @@ def __init__(self,
...
+            The group_method argument is deprecated and will be removed in the future.
+             Using it makes the dataloader have no length and prevents deterministic resumption.
+             It is also no longer needed, because concatenation can happen when we create
+             the dataset in create_c4.py.
+            """)
+        warnings.warn(DeprecationWarning(warning_msg))
```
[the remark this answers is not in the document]
dakinggg: and then we just immediately ban concat as an option? and also remove the --iter-- override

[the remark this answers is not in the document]
samhavens: it was decided to remove group_method

examples/common/convert_c4.py
samhavens: no worries

examples/common/text_data.py
```diff
@@ -63,12 +62,6 @@ def __init__(self,
                  num_canonical_nodes: Optional[int] = None,
                  batch_size: Optional[int] = None):
 
-        # Validation
```
dakinggg, line 66: Can we keep the deprecation warning you had, except make it a deprecation error? I believe you can do this by adding `kwargs` as a init arg, and then checking in kwargs for `group_method`

examples/common/text_data.py
```diff
@@ -96,78 +88,52 @@ def __init__(self,
...
+            # were added as part of pre-concatenating the text, so don't add them again
+            tokens = self.tokenizer(text_sample['text'],
+                                    truncation=False,
+                                    padding=False,
+                                    add_special_tokens=True)
+            if tokens['input_ids'][1] == self.tokenizer.bos_token_id or tokens[
+                    'input_ids'][-2] == self.tokenizer.eos_token_id:
+                self.add_special_tokens = False
+            else:
+                # Normal text without special tokens added
```
dakinggg, line 108:
```suggestion
                # Normal (not pre-concatenated) text without special tokens added
```

examples/common/text_data.py
```diff
@@ -96,78 +88,52 @@ def __init__(self,
...
+            np.frombuffer(sample['tokens'], dtype=np.int64).copy())
 
     # How to process a sample
     def __getitem__(self, idx: int) -> Dict[str, Any]:
-        text_sample = super().__getitem__(idx)
-        token_sample = self._tokenize(text_sample)
+        sample = super().__getitem__(idx)
+        if sample.get('text', None) is not None:
+            token_sample = self._tokenize(sample)
+        elif sample.get('tokens', None) is not None:
```
dakinggg, line 126: maybe add an else here that raises saying text or tokens must be present?

examples/common/text_data.py
```diff
@@ -218,9 +184,8 @@ def build_text_dataloader(cfg: DictConfig, device_batch_size: int):
             'remote': remote,
             'split': 'val',
             'shuffle': False,
-            'tokenizer_name': 'gpt2',
+            'tokenizer_name': 'facebook/opt-125m',
```
dakinggg, line 187: I think we should probably be consistent with which tokenizer we use in examples, so maybe stick with gpt2 here (and set the separator appropriately)?

examples/common/text_data.py
samhavens: Does python have a `DeprecationError`? Or are you saying that if I `raise DeprecationWarning('no group_method 4 u')` that will act like an error?

dakinggg: I think maybe just raise a `ValueError` or whatever other error you feel is appropriate, with the nice message that you wrote

examples/common/convert_c4.py
```diff
@@ -4,105 +4,392 @@
...
+        ]
 
-    Args:
-        samples (IterableDataset): An iterable dataset that is multi-worker compatible
+        if bos_text + eos_text == '':
+            test_tokens = tokenizer('test')
+            if test_tokens['input_ids'][
+                    0] != tokenizer.bos_token_id and test_tokens['input_ids'][
+                        -1] != tokenizer.eos_token_id:
+                tok_error_msg = 'This tokenizer does not insert an EOS nor BOS token.'
```
dakinggg, line 365:
```suggestion
                tok_error_msg = 'This tokenizer does not insert an EOS nor BOS token. '
```

examples/common/convert_c4.py
```diff
@@ -4,105 +4,392 @@
...
 
-        def __init__(self):
-            self.dataset = hf_datasets.load_dataset(path='c4',
-                                                    name='en',
-                                                    split=split,
-                                                    streaming=True)
+def _concat_sample_count(sample_count: int,
+                         mode: ConcatMode,
+                         max_length: Optional[int] = None) -> int:
+    emp_chars_per_sample = 2163  # measured for val split
```
dakinggg, line 327: what is the implication of these estimates? does it mean that we might end up returning from `generate_samples` early if these are an underestimate?

dakinggg (commented):
Mostly looks good to me. Left a few more comments. The other question I have is, do we want to merge this and then do convergence runs? or do convergence runs off of this branch before merging? cc @abhi-mosaic

examples/common/text_data.py
samhavens: luckily the later is effectively true in that it causes streaming to crash 🎉

examples/common/convert_c4.py
samhavens: yes

dakinggg: Is that a problem/why do we need to stop iteration at some sample number? can't we just exhaust the iterator?

samhavens: ohhh I don't need to do this, since I can keep track when building the buffers and concatenating...

dakinggg: ok yeah, lets do that. I still don't actually understand why we need these caps at all though.

requirements.txt
```diff
@@ -12,5 +12,6 @@ toml>=0.10.2,<0.11
 packaging>=21,<23
 omegaconf>=2.2.3,<3
 # these are just for common/text_data.py
+datasets>=2.9.0
 transformers>=4.11,<5
```
vchiley, line 16: The file header says 
> These requirements are shared for _**all**_ examples

Should 
```
datasets>=2.9.0
transformers>=4.11,<5
```
be part of this req file since they are not needed for any of the cv examples

examples/common/convert_c4.py
samhavens: only 3 for gpt2...

examples/common/convert_c4.py
samhavens: I think we only actually want it for `train_small` ?

requirements.txt
samhavens: hmmm... shared but only for text...

vchiley: **_ALL_**

lol.
yeah I guess this'll be @dblalock's call

examples/common/convert_c4.py
samhavens: `train` and `val` are the actual lengths. `train_small` is just an arbitrary amount. I made a change that uses the caps in a more reasonable way, only estimating to get tqdm a progress bar for the denominator

abhi-mosaic:
Made a pretty significant refactor to `convert_c4.py`, tokenization is a lot faster now: the `train` split with GPT2 is projected to take about 3hr on 64 CPU cores.

Some outstanding questions:
* make tokenization even faster using an inner loop within each CPU worker's `ConcatTokensC4`
* should we have an `allow_wrap: bool` flag? Right now with both concat options, the majority of written samples are "in media res", as in they are not starting exactly on the start of a sequence. Seems kinda bad? I think I would prefer throwing away the remainder of each concat_sample so that the starts are always true starts of real sequences. Or atleast have the option of trying that out.

abhi-mosaic:
FYI on how the parallelism works...

* we want to feed the `MDSWriter` with samples
* So we unpack individual samples from batches of a DataLoader
  * that DataLoader has many CPU workers
    * each CPU worker has its own copy of an `IterableDataset`, e.g. `ConcatTextC4`
      * that `ConcatTextC4` dataset has an internal `hf_dataset` with `streaming=True`
        * as of `datasets==2.9.0`, that `hf_dataset` is now smart enough to partition its shards
        * so that the `ConcatTextC4.hf_dataset` on CPU worker 0 has separate shards from `ConcatTextC4.hf_dataset` on CPU worker 1
      * Now we are defining `ConcatTextC4.__iter_`  to allow concatenation. 
      * So the samples are concatenated sequentially within a shard, within a CPU worker's partition of shards
    *  the CPU worker iterates over its `ConcatTextC4`, getting BS concatenated samples
  * the BS samples are handed back up to the DataLoader
  * by convention, each batch is fully produced by one CPU worker, when IterableDataset is used
  * so finally, the batch coming out of a DataLoader is BS concatenated sequences, all coming from one CPU worker
* and we unpack the samples to hand to MDSWriter

abhi-mosaic:
Added a param `--wrap` that (currently! not attached to this!) defaults to False, let me know what you think.

At a target character length of 10k, it creates a smaller `val` dataset of about 58574 vs. 78543 samples, or 75% of the size.

@samhavens @dakinggg curious what you think should be the default behavior. I imagine some kind of smarter bin-packing thing could help alleviate this drop, but not in scope for this PR...

## Default behavior:
```
$ python text_data.py concat_text_10k
Reading val split from concat_text_10k
num_samples: 58574


#################### Batch 0 ####################
input_ids torch.Size([2, 32]) torch.int64
attention_mask torch.Size([2, 32]) torch.int64
labels torch.Size([2, 32]) torch.int64
--------------------  Sample 0  --------------------
</s>The woman who died after falling from a bridge over the A21 has been identified as a Sevenoaks mum.
Marta Kendle,
--------------------  Sample 1  --------------------
</s>Beth has been writing short stories most of her life. She won third prize in the Women’s National Short Story Competition some years ago
```

## Using `--wrap`
```
$ python text_data.py concat_text_wrap_10k
Reading val split from concat_text_wrap_10k
num_samples: 78543


#################### Batch 0 ####################
input_ids torch.Size([2, 32]) torch.int64
attention_mask torch.Size([2, 32]) torch.int64
labels torch.Size([2, 32]) torch.int64
--------------------  Sample 0  --------------------
</s>The woman who died after falling from a bridge over the A21 has been identified as a Sevenoaks mum.
Marta Kendle,
--------------------  Sample 1  --------------------
l DG may be beneficial for the treatment of anxiety disorders. Copyright © 2013 Elsevier Inc. All rights reserved.
[1 ]Biology PhD Program,
```

dakinggg:
Thanks for figuring out the parallel stuff!

It is good we can experiment with both options. It seems like palm and gpt3 are both ok (if i'm reading right) with cutting examples in the middle. It does _seem_ like it might be better to not cut off examples, but I also don't love throwing away 25% of the data. This would also disproportionately throw away more text from longer samples I suspect, which seems like a bad bias to have. I'd be inclined to go for wrap true default, given the support from palm/gpt3.

palm:
```
Sequence length – A sequence length of 2048 was used for all models. Input examples are concatenated
together and then split into sequences of exactly 2048 tokens, so that there are no padding tokens, but
examples may be split in the middle. Input examples are differentiated from one another with a special
[eod] token.
```

t5:
```
We pre-train each model for 2**19 = 524,288 steps on C4 before fine-tuning. We use a
maximum sequence length of 512 and a batch size of 128 sequences. Whenever possible,
we “pack” multiple sequences into each entry of the batch so that our batches contain
roughly 2**16 = 65,536 tokens
```

gpt3:
```
During training we always train on sequences of the full nctx = 2048 token context window, packing multiple
documents into a single sequence when documents are shorter than 2048, in order to increase computational efficiency.
Sequences with multiple documents are not masked in any special way but instead documents within a sequence
are delimited with a special end of text token, giving the language model the information necessary to infer that
context separated by the end of text token is unrelated. This allows for efficient training without need for any special
sequence-specific masking.
```

examples/common/text_data.py
```diff
@@ -61,13 +60,14 @@ def __init__(self,
...
-        # Validation
-        if group_method not in ['truncate', 'concat']:
-            raise ValueError(
-                f"group_method='{group_method}' must be one of ['truncate', 'concat']."
-            )
+        if 'group_method' in kwargs:
+            raise DeprecationWarning(
+                'group_method is deprecated and has been removed.\nTo ' +
+                'concatenate, use the --concat_text or --concat_tokens ' +
+                'argument to concat_c4.py when creating your MDS dataset')
```
vchiley, line 70: python has deprecated warnings.
In python you can raise errors.
Can you raise a warning?

Is this valid python code? At the very least it seems like it [shouldn't](https://stackoverflow.com/a/29378596) be valid.

dakinggg: It is a...choice. Normally we would warn a `DeprecationWarning` and then later remove it, but in this case we _have_ to completely get rid of the option, because the previous code does not work with the latest streaming version. So this is a way for us to indicate that we deprecated something and explain it to the user, but also completely remove the option without a period of deprecation first. I'd be fine making it a `ValueError` or something though.

requirements.txt
dakinggg: they should probably be in both llm and bert, but with the same versions.

examples/common/text_data.py
vchiley: Maybe just say:
```
            raise NotImplementedError(
                'group_method is is no longer an option.\nTo ' +
                'concatenate, use the --concat_text or --concat_tokens ' +
                'argument to concat_c4.py when creating your MDS dataset'
```

dakinggg: Yeah, that works for me.

vchiley: We should also maybe check that there are not other kwargs that the user is trying to pass to the fn ie

```
kwargs.pop('group_method', None)  # remove the one possible kwarg
if kwargs:
    raise TypeError(f'StreamingTextDataset() got an unexpected keyword argument: {kwargs}')
```

vchiley (commented):
I read through [examples/common/text_data.py](https://github.com/mosaicml/examples/pull/128/files#diff-620e84172cafb01e42a31a23c94529acfccc2b096e224048f58fca5e4f34952a)
No major issues.
Comment on design: when we use it for pre-tokenized data is it still a "StreamingTextDataset" or should the classes be split up into "StreamingTextDataset" and "StreamingTokenDataset"?
Why do they need to be one class? There is a lot in the "StreamingTextDataset" that isn't explicitly needed in the "StreamingTokenDataset".

examples/common/convert_c4.py
```diff
@@ -4,105 +4,353 @@
...
+class EmptyOrNonexistentDirectory(Action):
+
+    def __call__(self, parser, namespace, values, option_string=None):
+        prospective_dir = values
+        if os.path.isdir(prospective_dir) and len(
+                os.listdir(prospective_dir)) > 0:
+            raise ArgumentError(self,
+                                f'--out_root={prospective_dir} must be empty')
+        else:
+            setattr(namespace, 'out_root', prospective_dir)
```
vchiley, line 36: I feel like most ppl just parse the args then check and raise a ValueError, or something, but this is fancy AF ❤️

examples/common/convert_c4.py
```diff
@@ -4,105 +4,353 @@
...
+        help='Convert text to tokens and concatenate up to this many tokens')
+
+    parser.add_argument('--tokenizer', type=str, required=False, default=None)
+    parser.add_argument('--bos_text', type=str, required=False, default=None)
+    parser.add_argument('--eos_text', type=str, required=False, default=None)
+    parser.add_argument('--wrap', default=False, action='store_true')
+
+    parsed = parser.parse_args()
+
+    if hasattr(parsed, 'concat_text') or hasattr(parsed, 'concat_tokens'):
```
vchiley, line 69: Why do the following two if statements need to be guarded by this if statement?

examples/common/convert_c4.py
abhi-mosaic: Also, I think this is a bit too strict, like if you write the `val` split you should still be allowed to later write the `train` split?

requirements.txt
dakinggg: i'm going to leave them here for now and create a task to fix it. fixing it means also gating the transformers/datasets imports in common.

examples/common/convert_c4.py
dakinggg: had to remove this beauty to let it be less strict :P

examples/common/convert_c4.py
dakinggg: it doesn't. removed.

samhavens: at one point I was using a subparser for `concat` and so it was possible for these to not be attributes of `parsed`. They always exist now so I will remove this guard

examples/common/convert_c4.py
samhavens: Abhi is right, but I have to remove the fancy `Action` subclass, since it can't validate against other args (we need to know splits) since it hasn't parsed that yet :sob:

samhavens: @dakinggg  beat me to it

samhavens:
I agree with @dakinggg. "This would also disproportionately throw away more text from longer samples I suspect, which seems like a bad bias to have. I'd be inclined to go for wrap true default, given the support from palm/gpt3." We could experiment with both and see if it affects convergence... I have also always wondered about adding a MOS token to tell the model that we are starting without needed context, and then just never showing it that at inference time.

examples/common/text_data.py
```diff
@@ -179,6 +164,7 @@ def build_text_dataloader(cfg: DictConfig, device_batch_size: int):
         shuffle_seed=cfg.dataset.get('shuffle_seed', None),
         num_canonical_nodes=cfg.dataset.get('num_canonical_nodes', 128),
         batch_size=device_batch_size)
+    print(f'num_samples: {len(dataset)}')
```
abhi-mosaic, line 167:
```suggestion
```
This was a bad idea on my part, because len(dataset) will be 1/world_size. Would rather not import `dist` here unnecessarily, so recommending removal.

examples/llm/README.md
```diff
@@ -78,12 +78,15 @@ To make yourself a copy of C4, use `convert_c4.py` like so:
 # Download the 'train_small' and 'val' splits and convert to StreamingDataset format
 # This will take 20-60 seconds depending on your Internet bandwidth
 # You should see two folders: `./my-copy-c4/train_small` and `./my-copy-c4/val` that are each ~0.5GB
-python ../common/convert_c4.py --out_root ./my-copy-c4 --splits train_small val
+# Note: We are using the `--concat_tokens` option to pre tokenize our samples to be of the max sequence length without padding
+# `--concat_text` is also an option if you don't want to pre tokenize, but may result in some padding or truncated text in your samples,
+# which affects the MFU calculation
+python ../common/convert_c4.py --out_root ./my-copy-c4 --splits train_small val --concat_tokens 2048 --tokenizer gpt2 --bos_text '<|endoftext|>'
```
abhi-mosaic, line 84: I don't think we want `bos_text` here, we want `eos_text`.

dakinggg: oh yeah, woops

examples/common/text_data.py
```diff
@@ -202,25 +187,30 @@ def build_text_dataloader(cfg: DictConfig, device_batch_size: int):
...
+    parser.add_argument("--remote_path", type=str, default=None, help='the path to the remote copy to stream from (optional)')
+    parser.add_argument("--split", type=str, default="val", help='which split of the dataset to use')
+
+    args = parser.parse_args()
 
-    if len(sys.argv) > 2:
-        local, remote = sys.argv[1:3]
-        print(f'Reading val split from {local} <- streamed from <- {remote}')
+    if args.remote_path is not None:
+        print(f'Reading val split from {args.local_path} <- streamed from <- {remote_path}')
```
abhi-mosaic, line 201:
```suggestion
        print(f'Reading val split from {args.local_path} <- streamed from <- {args.remote_path}')
```

dakinggg:
Ok, I think this is good to merge, not sure who is approving it at this point since we've all contributed, but @abhi-mosaic and @samhavens have another look tomorrow and then we can merge.

examples/bert/README.md
```diff
@@ -79,12 +82,12 @@ To verify that the dataloader works, run a quick test on your `val` split like s
...
 # Since we only provide a local path, no streaming/copying takes place.
-python ../common/text_data.py ./my-copy-c4
+python ../common/text_data.py --local_path ./my-copy-c4 --tokenizer bert-base-uncased
 
 # This will do the same thing, but stream data to {local} from {remote}.
 # The remote path can be a filesystem or object store URI.
-python ../common/text_data.py /tmp/cache-c4 ./my-copy-c4  # stream from filesystem, e.g. a slow NFS volume to fast local disk
-python ../common/text_data.py /tmp/cache-c4 s3://my-bucket/my-copy-c4  # stream from object store
+python ../common/text_data.py --local_path /tmp/cache-c4 --remote_path ./my-copy-c4 --tokenizer bert-base-uncased # stream from filesystem, e.g. a slow NFS volume to fast local disk
+python ../common/text_data.py --local_path /tmp/cache-c4 --remote_path s3://my-bucket/my-copy-c4 --tokenizer bert-base-uncased # stream from object store
```
vchiley, line 90: in `llm` repo, this equivalent line is commented out.
we should pick if we want to comment this one out and stay consistent between the two readme

dakinggg: yup, will comment out here.

examples/common/text_data.py
```diff
@@ -61,13 +60,18 @@ def __init__(self,
...
-            )
+                 batch_size: Optional[int] = None,
+                 **kwargs: Dict[str, Any]):
+
+        group_method = kwargs.pop('group_method', None)
+        if group_method is not None:
+            raise NotImplementedError(
+                'group_method is deprecated and has been removed.\nTo ' +
+                'concatenate, use the --concat_text or --concat_tokens ' +
+                'argument to concat_c4.py when creating your MDS dataset')
```
vchiley, line 71:
```suggestion
                'concatenate, pass the --concat_text or --concat_tokens ' +
                'argument to concat_c4.py when creating your MDS dataset')
```
or
```suggestion
                'concatenate, use the --concat_text or --concat_tokens ' +
                'argument when creating your MDS dataset with concat_c4.py')
```

examples/common/text_data.py
```diff
@@ -84,81 +88,63 @@ def __init__(self,
                          batch_size=batch_size)
         self.tokenizer_name = tokenizer_name
         self.max_seq_len = max_seq_len
-        self.group_method = group_method
 
         # Build tokenizer
         os.environ['TRANSFORMERS_NO_ADVISORY_WARNINGS'] = '1'
+        os.environ['TOKENIZERS_PARALLELISM'] = 'false'
         self.tokenizer = transformers.AutoTokenizer.from_pretrained(
             self.tokenizer_name)
```
vchiley, line 96:
```suggestion
        self.tokenizer = transformers.AutoTokenizer.from_pretrained(
            tokenizer_name)
```
Not sure why we set `self.tokenizer_name = tokenizer_name` in line 89.

examples/common/text_data.py
```diff
@@ -84,81 +88,63 @@ def __init__(self,
...
-        if self.tokenizer._pad_token is None:
-            # Some tokenizers (e.g. GPT2 tokenizer) have no padding token which causes bugs
-            self.tokenizer._pad_token = self.tokenizer._eos_token
-        # suppress warnings when using group_method='concat' and no truncation
+
+        # suppress warnings when using longer 'max_seq_len'
         self.tokenizer.model_max_length = int(1e30)
```
vchiley, line 99: is there a reason to set this to `1e30`? instead of using `max_seq_len`

examples/common/text_data.py
```diff
@@ -167,7 +153,6 @@ def build_text_dataloader(cfg: DictConfig, device_batch_size: int):
         local=cfg.dataset.local,
         tokenizer_name=cfg.dataset.tokenizer_name,
         max_seq_len=cfg.dataset.max_seq_len,
-        group_method=cfg.dataset.group_method,
         remote=cfg.dataset.get('remote', None),
         split=cfg.dataset.get('split', None),
         shuffle=cfg.dataset.get('shuffle', False),
```
vchiley, line 158: `keep_zip` default should be `None` in keeping with the default [here](https://streaming.docs.mosaicml.com/en/stable/api_reference/generated/streaming.StreamingDataset.html#streamingdataset) and [here](https://github.com/samhavens/benchmarks/blob/R-387--update-scripts-to-pre-concat/examples/common/text_data.py#L57)

examples/llm/README.md
samhavens: Honestly, if I was going to do one, I'd use it as BOS. Since we are just picking one, it won't be any different at the boundary between two sequences. The difference is only for the first segment in the sequence. And at the beginning of the sequence, I think it will just have to use absolute unigram probabilities for the first token, rather than the distribution given that it is the first token in the sequence (correct me if I'm wrong there). Here is an n=1 experiment in favor of this POV https://github.com/huggingface/transformers/issues/3311

examples/common/text_data.py
```diff
@@ -202,25 +187,30 @@ def build_text_dataloader(cfg: DictConfig, device_batch_size: int):
 # Helpful to test if your dataloader is working locally
 # Run `python data.py [remote] [local, optional]` and verify that batches are printed out
```
vchiley, line 188:
```suggestion
# Run `python data.py  --local_path [local] [--remote_path remote, optional]` and verify that batches are printed out
```

examples/common/text_data.py
```diff
@@ -202,25 +187,30 @@ def build_text_dataloader(cfg: DictConfig, device_batch_size: int):
...
 if __name__ == '__main__':
+    import argparse
+
+    parser = argparse.ArgumentParser()
+    parser.add_argument("--tokenizer", type=str, default="gpt2", help="the name of the tokenizer to use")
+    parser.add_argument("--local_path", type=str, required=True, help='the path to the local copy of the dataset')
+    parser.add_argument("--remote_path", type=str, default=None, help='the path to the remote copy to stream from (optional)')
+    parser.add_argument("--split", type=str, default="val", help='which split of the dataset to use')
+
```
vchiley, line 197:
```suggestion
    parser.add_argument("--max_seq_len", type=int, default=32, help='max sequence length to test')

```

examples/common/text_data.py
```diff
@@ -202,25 +187,30 @@ def build_text_dataloader(cfg: DictConfig, device_batch_size: int):
...
-            'local': local,
-            'remote': remote,
-            'split': 'val',
+            'local': args.local_path,
+            'remote': args.remote_path,
+            'split': args.split,
             'shuffle': False,
-            'tokenizer_name': 'gpt2',
+            'tokenizer_name': args.tokenizer,
             'max_seq_len': 32,
```
vchiley, line 213:
```suggestion
            'max_seq_len': args.max_seq_len,
```

examples/common/text_data.py
```diff
@@ -84,81 +88,63 @@ def __init__(self,
...
+                    'input_ids'][-2] == self.tokenizer.eos_token_id:
+                self.add_special_tokens = False
+            else:
+                # Normal (not pre-concatenated) text without special tokens added
+                self.add_special_tokens = True
+
+        if self.tokenizer._pad_token is None:
+            # Some tokenizers (e.g. GPT2 tokenizer) have no padding token which causes bugs
+            raise RuntimeError(
+                'If tokenizing on-the-fly, tokenizer must have a pad_token_id')
```
vchiley, line 123: Should we enable users to set a `_pad_token`?
either as a kwarg or provide a member method to our class?

examples/common/convert_c4.py
```diff
@@ -4,105 +4,344 @@
...
+    parser.add_argument('--compression', type=str, default=None)
+
+    group = parser.add_mutually_exclusive_group(required=False)
+    group.add_argument('--concat_text',
+                       type=int,
+                       help='Number of characters to concatenate to')
+    group.add_argument(
+        '--concat_tokens',
+        type=int,
+        help='Convert text to tokens and concatenate up to this many tokens')
```
vchiley, line 45: it looks weird that the formatting is different.
it looks like a pre-commit linter fixed it and we didn't catch it 😅

samhavens: I think this is "intentional" with our formatting penalties. It tries to keep some of the args on the first line and align with them, but that would cause the second help text to go past a threshold?

examples/common/convert_c4.py
```diff
@@ -4,105 +4,344 @@
...
+            while len(buffer) >= self.max_length:
+                concat_sample = buffer[:self.max_length]
+                buffer = buffer[self.max_length:] if self.should_wrap else []
+                yield {
+                    # convert to bytes to store in MDS binary format
+                    'tokens': np.asarray(concat_sample).tobytes()
+                }
+
+
+def build_hf_c4_dataset(
```
vchiley, line 207: most of `_get_kwargs` output is used by `build_hf_c4_dataset`.
`build_hf_c4_dataset` should absorb `_get_kwargs`.

examples/common/convert_c4.py
```diff
@@ -4,105 +4,344 @@
...
+        self,
+        split: str,
+        tokenizer: PreTrainedTokenizerBase,
+        max_length: int,
+        bos_text: str,
+        eos_text: str,
+        no_wrap: bool,
+    ):
+        self.tokenizer = tokenizer
+        os.environ['TOKENIZERS_PARALLELISM'] = 'false'
```
vchiley, line 161: Should this be set in `NoConcatC4` and `ConcatTextC4` as well?

examples/common/convert_c4.py
```diff
@@ -4,105 +4,344 @@
...
+        for sample in self.hf_dataset:
+            buffer = buffer + self.bos_text + sample['text'] + self.eos_text
+            while len(buffer) >= self.max_length:
+                concat_sample = buffer[:self.max_length]
+                buffer = buffer[self.max_length:] if self.should_wrap else ''
+                # convert to bytes to store in MDS binary format
+                yield {'text': concat_sample.encode('utf-8')}
+
+
+class ConcatTokensC4(IterableDataset):
```
vchiley, line 128: Should inherit from `ConcatTextC4`.

examples/common/convert_c4.py
```diff
@@ -4,105 +4,344 @@
...
-    return args.parse_args()
+    def __iter__(self) -> Iterable[Dict[str, bytes]]:
+        for sample in self.hf_dataset:
+            # convert to bytes to store in MDS binary format
+            yield {'text': sample['text'].encode('utf-8')}
 
 
-def build_hf_c4_dataset(split: str) -> IterableDataset:
-    """Collect the samples for this dataset split.
+class ConcatTextC4(IterableDataset):
```
vchiley, line 100: Should inherit from `NoConcatC4`.

examples/common/convert_c4.py
```diff
@@ -4,105 +4,344 @@
...
 from streaming import MDSWriter
 from torch.utils.data import DataLoader, IterableDataset, get_worker_info
 from tqdm import tqdm
+from transformers import AutoTokenizer, PreTrainedTokenizerBase
+
+
+class ConcatMode(Enum):
+    NO_CONCAT = 'NO_CONCAT'
+    CONCAT_TEXT = 'CONCAT_TEXT'
+    CONCAT_TOKENS = 'CONCAT_TOKENS'
```
vchiley, line 24: Is this needed? why not just let `args.concat_tokens` and `args.concat_text` handle this?

samhavens: The mode is derivable from those. The hope was this made it clearer, but I can remove

examples/common/text_data.py
samhavens: Since this class is instantiated from configs, I think we should do it as a kwarg. Sound good?

examples/common/text_data.py
samhavens: I don't think so. It was in here because of on the fly concat.

examples/llm/README.md
samhavens: @abhi-mosaic thoughts on me reverting to bos?

examples/common/text_data.py
samhavens: @abhi-mosaic said "Now that we have streaming==0.2.2, the default for num_canonical_nodes should probably be made high (128 or 256 or so) to establish good shuffling randomness." Not sure if we should set it to 128 everywhere now...?

examples/common/text_data.py
dakinggg: @abhi-mosaic said removing it caused an error, so I propose leaving it for now, rather than investigating.

samhavens:
Request based on some customer issues:

If `data_remot`e is `None` or is equal to `data_local`, instead of letting streaming attempt to build the dataset and raising a `FileNotFound` error about the `index.json`, before even instantiating the Dataset, check if the `split` directory is present in the `data_local` directory. If not, raise a `ValueError` that the split isn't present

examples/llm/README.md
dakinggg: I am unopinionated, although gpt paper does call it an EOS, not a BOS

examples/common/text_data.py
abhi-mosaic: The 128 is quite important for good shuffling, and I did some testing to pick a value that did not add too much overhead on init (<15s): https://github.com/mosaicml/examples/pull/123#issue-1573524320

It is probably a good idea for Streaming team to set a default large number too... I will make a JIRA for them.

The other defaults, I guess it's fine to copy them over? Is it also a good idea to put the 128 value in the default of our `StreamingText` class? Maybe

examples/common/convert_c4.py
abhi-mosaic: I think there are no tokenizers there, so not needed

examples/common/convert_c4.py
abhi-mosaic: I really don't think the inheritance does much here... like if I'm an external person and I just want the code for concatenating tokens, easy to just rip this class. Rather than rip 3 classes

dakinggg:
Added the extra checking of split in local dir

examples/common/convert_c4.py
dakinggg: I think the mode makes it clearer, so will leave it.

examples/common/convert_c4.py
dakinggg: yeah, i guess you'd get a few items shared in the constructor, but that is it. I think I will agree with abhi and leave as is for now.

examples/common/convert_c4.py
dakinggg: agreed, will do

examples/common/text_data.py
dakinggg: I think we actually don't need to add this. We pass along tokenizer kwargs, so the user can just add `pad_token: <PAD>` to their yaml, and they will end up with a tokenizer with a pad token

dakinggg: oh maybe we don't pass along all the kwargs...checking

dakinggg: ok we don't pass along the kwargs, but we do still automatically override pad with eos if pad is none. that is in `tokenizer.py`. Not sure if we should get rid of that or not...

examples/common/text_data.py
dakinggg: set `keep_zip` to `None`, defaulted `num_canonical_nodes` to 128 everywhere, change `shuffle_seed` default to 9176.

examples/common/text_data.py
dakinggg: deferring this to another PR, but planning still before release.

examples/common/convert_c4.py
dakinggg: ok refactored. still testing it.

abhi-mosaic (approved):
Been following this PR for a while, and the recent changes don't appear to break anything. Punting `--concat_text` to the next release (or whenever we actually use it) is the smart move. 

Let's merge it!

PR #129, opened 2023-02-07
Clarify BertConfig choices for BERT yamls and switch fused_layer_norm to low_precision_layernorm

jacobfulano:
Our default yamls don't make it clear how basic model architecture configuration choices are made (such as the number of attention heads).

Additionally, this makes clear how a user might define the ALiBi mask size for different sequence lengths.

```
   # Mosaic BERT 'base' generally uses the default architecture values for from the Hugging Face BertConfig object
    # These values can be changed here:
    model_config:
      num_attention_heads: 12 # bert-base default
      num_hidden_layers: 12 # bert-base default
      max_position_embedding: 512 # This is used for defining the shape of the ALiBi mask
      attention_probs_dropout_prob: 0.0 # This must be 0 for Flash Attention with triton to work
```

I think this is the easiest way of dealing with the question of max sequence length. In our current implementation for BERT, the ALiBi mask is initiated with size 512 (using the default of `max_position_embedding` from BertConfig). This causes issues for using a maximum sequence length >512.

I'm open to other suggestions. 

An alternative is to use:
```
max_position_embedding: *max_seq_len
```
Although I don't fully understand the implications of training with a small initialized ALiBi mask and testing on longer sequence lengths.


examples/bert/yamls/main/hf-bert-base-uncased.yaml
```diff
@@ -19,6 +19,13 @@ model:
   use_pretrained: false # Train the model from scratch. Set to true to start from the HF off-the-shelf weights.
   pretrained_model_name: *tokenizer_name
   tokenizer_name: *tokenizer_name
+  # This implementation generally uses the default architecture values for from the Hugging Face BertConfig object
+  # These values can be changed here:
+  model_config:
```
alextrott16, line 24: For the HF example, which comes with a `use_pretrained` option, we may want to clarify that config changes that change the size of the model (compared to the defaults registered for `pretrained_model_name`) will create issues when using `use_pretrained: true`.

alextrott16:
I think we made a poor design decision in having MosaicBERT create an Alibi _buffer_ that uses `config.max_position_embeddings` to set its shape.

1) It creates a different meaning for that config variable, which is confusing.
2) We end up using it to define a buffer with a fixed size. Because it's a buffer, it gets saved into the model checkpoint and when you want to load that checkpoint, you need the buffer of the downstream model to have the same size as what's in the checkpoint. In other words, it becomes needlessly tedious to do fine-tuning/inference on sequences larger than the `config.max_position_embeddings` used during pre-training.

It might be outside the scope of this PR, but in order to better clarify the BERT config settings, we should really just have MosaicBERT ignore the `max_position_embeddings` config field (since we have removed position embeddings) and make the alibi biases a dynamic tensor that is not registered as a buffer.

@jacobfulano What do you think? I can take a stab at those changes if you think it's a good idea.

examples/bert/yamls/main/hf-bert-base-uncased.yaml
jacobfulano: Agreed that this is worth including

jacobfulano:
> I think we made a poor design decision in having MosaicBERT create an Alibi _buffer_ that uses `config.max_position_embeddings` to set its shape.
> 
> 1. It creates a different meaning for that config variable, which is confusing.
> 2. We end up using it to define a buffer with a fixed size. Because it's a buffer, it gets saved into the model checkpoint and when you want to load that checkpoint, you need the buffer of the downstream model to have the same size as what's in the checkpoint. In other words, it becomes needlessly tedious to do fine-tuning/inference on sequences larger than the `config.max_position_embeddings` used during pre-training.

These two things tripped me up! I agree that they add some extra potential pain points for users.

> It might be outside the scope of this PR, but in order to better clarify the BERT config settings, we should really just have MosaicBERT ignore the `max_position_embeddings` config field (since we have removed position embeddings) and make the alibi biases a dynamic tensor that is not registered as a buffer.
> 
> @jacobfulano What do you think? I can take a stab at those changes if you think it's a good idea.

I agree with this. @alextrott16 feel free to take a stab at this separate from this PR (which is currently just small yaml changes)

alextrott16:
@jacobfulano OK, maybe it makes the most sense to pause this PR until I have the new alibi handling in place (should be doable by tomorrow). Then you'll have a clearer picture for what text is most appropriate in the YAMLs.

examples/bert/yamls/main/hf-bert-base-uncased.yaml
jacobfulano: I've added an extra line in the descriptions
```
# These values can be changed here when pretraining from scratch. Note that these should only be used
  # if used_pretained: false, otherwise the model will not be loaded properly
  ```

**Comment by jacobfulano** [comment]
@alextrott16 I made the extra config very simple, with some comments that just make the number of layers explicit.

I also switched to low_precision_layernorm here.

After linting, there are a few extra files that have superficial spacing changes

**Review by alextrott16 — CHANGES_REQUESTED** [review]
```
# These values can be changed here when pretraining from scratch. Note that these should only be used
# if used_pretained: false, otherwise the model will not be loaded properly
```
This text works for the HF examples, but MosaicBERT needs slightly different language. In MosaicBERT, there is not a `use_pretrained` argument but instead a `pretrained_checkpoint` argument. For MosaicBERT, I think we want to say something like "Note: if using the `pretrained_checkpoint` argument to create a model from an existing checkpoint, make sure the `model_config` settings match the architecture of the existing model".

Otherwise, LGTM! Happy to approve once the language is updated.

jacobfulano:
Just updated! Thanks for this

alextrott16 (approved):

PR #130, opened 2023-02-07
Add reproducibility.seed_all, set pip install, and fix cd command

Landanjs:
Without reproducibility.seed_all, streaming will freeze. Pip install is needed after the recent refactor


A-Jacobson (approved):
assuming the freezing bug is getting fixed in streaming at some point. 

since  I should have added a seed anyway. LGTM.

PR #131, opened 2023-02-08
lower dataset size back down and lower canonical nodes

dakinggg:
- change the bert test dataset size back down to 16 by setting `num_canonical_nodes` smaller
- move flash attention import so the file can still be imported on cpu. verified that llm example can still begin training.


abhi-mosaic (approved):
LGTM pending passing checks.

PR #132, opened 2023-02-08
De-buffer MosaicBERT alibi

alextrott16:
Addresses checkpointing/loading complications and weird config naming convention related to ALiBi.

Alters MosaicBERT so that (1)`alibi` is an attribute of the model instead of a registered buffer, and (2) the alibi tensor is rebuilt as needed and is sized according to the longest sequence seen since model initialization.

### DIFF
--- a/examples/bert/src/bert_layers.py
+++ b/examples/bert/src/bert_layers.py
@@ -409,13 +409,17 @@ class BertEncoder(nn.Module):
 
         self.num_attention_heads = config.num_attention_heads
 
-        self.alibi: Optional[torch.Tensor] = None
-        self._current_alibi_size = -1
+        # The alibi mask will be dynamically expanded if it is too small for
+        # the input the model receives. But it generally helps to initialize it
+        # to a reasonably large size to help pre-allocate CUDA memory.
+        # We initialize it here to handle any sequences up to 512 tokens.
+        self.alibi = self.build_alibi_tensor(self.num_attention_heads, size=512)
+        self._current_alibi_size = 512
 
     @staticmethod
     def build_alibi_tensor(
             n_heads: int,
-            max_token_length: int,
+            size: int,
             device: Optional[Union[torch.device, str]] = None) -> torch.Tensor:
         # Alibi
         # Following https://github.com/ofirpress/attention_with_linear_biases/issues/5 (Implementation 1)
@@ -435,16 +439,15 @@ class BertEncoder(nn.Module):
             # workaround.
             if math.log2(n_heads).is_integer():
                 return get_slopes_power_of_2(n_heads)
-            else:
-                closest_power_of_2 = 2**math.floor(math.log2(n_heads))
-                return get_slopes_power_of_2(
-                    closest_power_of_2) + _get_alibi_head_slopes(
-                        2 * closest_power_of_2)[0::2][:n_heads -
-                                                      closest_power_of_2]
 
-        context_position = torch.arange(max_token_length, device=device)[:,
-                                                                         None]
-        memory_position = torch.arange(max_token_length, device=device)[None, :]
+            closest_power_of_2 = 2**math.floor(math.log2(n_heads))
+            slopes_a = get_slopes_power_of_2(closest_power_of_2)
+            slopes_b = _get_alibi_head_slopes(2 * closest_power_of_2)
+            slopes_b = slopes_b[0::2][:n_heads - closest_power_of_2]
+            return slopes_a + slopes_b
+
+        context_position = torch.arange(size, device=device)[:, None]
+        memory_position = torch.arange(size, device=device)[None, :]
         relative_position = torch.abs(memory_position - context_position)
         # [n_heads, max_token_length, max_token_length]
         relative_position = relative_position.unsqueeze(0).expand(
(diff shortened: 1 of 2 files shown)


examples/bert/src/bert_layers.py
```diff
@@ -474,7 +481,18 @@ def forward(
...
         # Add alibi matrix to extended_attention_mask
+        if self._current_alibi_size < seqlen or self.alibi is None:
+            # Build/rebuild the alibi tensor if needed
+            self.alibi = self.build_alibi_tensor(
+                n_heads=self.num_attention_heads,
+                max_token_length=seqlen,
+                device=hidden_states.device)
+            self._current_alibi_size = int(seqlen)
+        if self.alibi.device != hidden_states.device:
+            # Device catch-up
```
jacobfulano, line 494: For a noob like me, can you clarify that this makes sure that the alibi mask is synced with the correct device(s)? Why wouldn't it be otherwise?

examples/bert/src/bert_layers.py
```diff
@@ -474,7 +481,18 @@ def forward(
...
+            self.alibi = self.alibi.to(hidden_states.device)
         alibi_bias = self.alibi[:, :, :seqlen, :seqlen]  # type: ignore
         alibi_attn_mask = extended_attention_mask[:, :, :, :seqlen] + alibi_bias
```
jacobfulano, line 498: This is great! I like that the ALiBi tensor is built to the right size on the fly. This seems like the correct way to do it

examples/bert/src/bert_layers.py
```diff
@@ -434,21 +442,20 @@ def get_slopes_power_of_2(n_heads: int) -> List[float]:
                         2 * closest_power_of_2)[0::2][:n_heads -
                                                       closest_power_of_2]
 
-        max_token_length = config.max_position_embeddings
-        context_position = torch.arange(max_token_length)[:, None]
-        memory_position = torch.arange(max_token_length)[None, :]
+        context_position = torch.arange(max_token_length, device=device)[:,
+                                                                         None]
```
jacobfulano, line 446: Is the extra line here intentional?

examples/bert/src/bert_layers.py
jacobfulano: It also means that the mcloud configs don't have to deal with anything other than the maximum sequence length =)

examples/bert/src/bert_layers.py
alextrott16: Yes, that is what's happening! We're assuming that the inputs to the model are on the correct device, so we're using `hidden_states.device` to get that device.

The alibi mask might not be on the correct device if something caused `self.alibi` to be created (i.e. the model's forward method was called) in between the model being on CPU and the model being moved to another device. In that case, `self.alibi` would still be on CPU and the device would need to "catch up" with the rest of the model :)

examples/bert/src/bert_layers.py
alextrott16: Sigh. No, it's the auto formatter being a bully. I might try to see if I can outsmart it because this is nuts.

alextrott16: Aha! Problem solved.

vchiley: > auto formatter

BOOOOOOOOO

examples/bert/src/bert_layers.py
```diff
@@ -474,7 +483,18 @@ def forward(
...
+            self.alibi = self.build_alibi_tensor(
+                n_heads=self.num_attention_heads,
+                size=seqlen,
+                device=hidden_states.device)
+            self._current_alibi_size = int(seqlen)
+        elif self.alibi.device != hidden_states.device:
+            # Device catch-up
+            self.alibi = self.alibi.to(hidden_states.device)
         alibi_bias = self.alibi[:, :, :seqlen, :seqlen]  # type: ignore
         alibi_attn_mask = extended_attention_mask[:, :, :, :seqlen] + alibi_bias
```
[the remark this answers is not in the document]
vchiley: I'm technically future proofing for if extended_attention_mask is ever extended to be have shape [b, h, seqlen, seqlen] since currently it has shape [b, 1, 1, seqlen] but even if shape is [b, 1, 1, seqlen], `[:, :, :seqlen, :seqlen]` should work because python / pytorch is smart that way (at least I've done it in the past to tensors which could have either shape).
Given that this is a "future proofing" feel free to ignore this suggestion.

jacobfulano:
Happy to approve before bug bash

examples/bert/src/bert_layers.py
```diff
@@ -474,7 +483,18 @@ def forward(
         #        hidden_states[ntokens,hidden] -> hidden_states[ntokens_unpad,hidden]
         hidden_states, indices, cu_seqlens, _ = unpad_input(
             hidden_states, attention_mask_bool)
+
         # Add alibi matrix to extended_attention_mask
+        if self._current_alibi_size < seqlen:
+            # Build/rebuild the alibi tensor if needed
```
vchiley, line 489: Should we add a warning every time we do this? Something like:
warnings.warn(f'Increasing max alibi size from {self._current_alibi_size} to {seqlen}')

examples/bert/src/bert_layers.py
```diff
@@ -409,6 +409,18 @@ def __init__(self, config):
 
         self.num_attention_heads = config.num_attention_heads
 
+        # The alibi mask will be dynamically expanded if it is too small for
+        # the input the model receives. But it generally helps to initialize it
+        # to a reasonably large size to help pre-allocate CUDA memory.
+        # We initialize it here to handle any sequences up to 512 tokens.
+        self.alibi = self.build_alibi_tensor(self.num_attention_heads, size=512)
+        self._current_alibi_size = 512
```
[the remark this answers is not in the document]
alextrott16: I was trying to prioritize use of static methods, but in considering your question I think that was silly. Will update.

PR #133, opened 2023-02-08
fix deeplab requirements

dakinggg:
`mmcv-full` requires a command specific to torch/cuda version, so I don't think can be included in `requirements.txt`


dblalock (approved):
LGTM

PR #135, opened 2023-02-08
README cleanup + consistent-ization

dblalock:
Went through the CIFAR, ResNet, DeepLab, and BERT READMEs and:

1. Added mention of how to train with mcloud
2. Made them a bit more consistent
3. Made only the example name H1, with everything else H2 or less
4. Tweaked prose a little
5. Ripped out the 500-word essay on why should pretrain from the BERT readme. It's good content but this readme is 2700 words even without it so we really need to cut length.

BERT is still kind of the odd one out with a different structure and a *way* longer. We need to think carefully about how to structure this one to handle the cartesian product of {pretrain, finetune} {local, mcloud}.

To review, I would recommend previewing the resulting markdown real fast.


dblalock:
Added several people as reviewers in the hopes that one of you will have bandwidth before the start of the bug bash

README.md
```diff
@@ -20,7 +20,9 @@ pip install -e ".[llm]"  # or pip install -e ".[llm-cpu]" if no NVIDIA GPU
 cd examples/llm
 ```
 
-# Extending an example
+Available examples include `bert`, `cifar`, `llm`, and `resnet`.
```
dakinggg, line 23: deeplab and nemo?

dakinggg (approved):
quick skim LGTM

PR #138, opened 2023-02-09
Update Nemo Megatron

bcui19:
This change updates the nemo example as follows (from PR #115):

- Fixed a bug in the seq command which launched one process too many, failing the job
- Update to run_name to have an informative default
- Updating base image to latest nemo image 22.11 
- Added comments better document the setup 
- Moved apt install into the integrations section
- Updated a few hyper parameters to ensure successful run

Both YAMLs were tested and ran successfully on r7z2 (tested with @bcui19 's pre-processed data stored at s3://mosaicml-internal-checkpoints-shared/brandon/NeMo/)

Also removed linting issues.


dakinggg (approved):
LGTM

PR #139, opened 2023-02-09
add nemo to linting ci

dakinggg:


dblalock (approved):
I see it in the CI below, so LGTM if it passes

PR #140, opened 2023-02-09
fsdp for bert

jacobfulano:
This is a draft PR for FSDP implementation in BERT

### DIFF
--- a/examples/bert/src/bert_layers.py
+++ b/examples/bert/src/bert_layers.py
@@ -37,6 +37,7 @@ import copy
 import logging
 import math
 import warnings
+from functools import partial
 from typing import List, Optional, Tuple, Union
 
 import torch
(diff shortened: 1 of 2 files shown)


examples/bert/src/bert_layers.py
```diff
@@ -545,6 +546,70 @@ def forward(
             all_encoder_layers.append(hidden_states)
         return all_encoder_layers
 
+    # Param Initialization, needed for device='meta' fast initialization
+    def param_init_fn(self, module):
```
abhi-mosaic, line 550: @jacobfulano a few comments:
* `param_init_fn` is not needed unless you plan to make your models `meta` initializable. For a first draft, I would remove entirely
* `fsdp_wrap_fn` and `activation_checkpointing_fn` need to be defined on a root module underneath the `ComposerModel`. So they have to be defined for say, `BertForMaskedLM`. I would recommend moving the definition there, and I think it will work!

PR #141, opened 2023-02-09
Add starter script for single-task classification fine-tuning

alextrott16:
There is currently a blind-spot in the `bert` folder when it comes to fine-tuning. The GLUE script is too complicated to reflect a typical use-case, so I added `classification.py` as a simple starter script to make it easier for someone to fine-tune a BERT model on some classification task.

In addition to the new script, I have modified the README, added configs, and added a test for the new script.


examples/bert/README.md
```diff
@@ -129,14 +131,24 @@ composer main.py yamls/test/main.yaml model.name=mosaic_bert
...
 
 ```bash
 # Run the fine-tuning script with the test config and HuggingFace BERT
+python classification.py yamls/test/classification.yaml
+
+# Run the fine-tuning script with the test config and Mosaic BERT
+python classification.py yamls/test/classification.yaml model.name=mosaic_bert
+```
+
+Second, very `glue.py` for both models.
```
samhavens, line 144: should this be "verify"?

samhavens (commented):
I think everyone assumes that "classification" means "sequence classification", but if we ever add a token classification example, maybe we'll wish we'd called this "sequence_classification" instead?

jacobfulano:
This is very helpful! Code lgtm and is a pleasure to read

jacobfulano (approved):
Clean and easy to read.

PR #142, opened 2023-02-09
Update deeplab

dakinggg:
- pins mmseg to 0.22.0 (the version in our docker image)
- mentions that our docker image uses mmcv-full 1.4.8
- fixes an import path


dblalock (approved):
LGTM, though would be great to have someone else test this

PR #145, opened 2023-02-09
Handle expected None case for GPTNeoX

alextrott16:
Was seeing errors from `tied_embedding` module being `None` with GPTNeoX. But, as mentioned in the comments, that is expected. So, I made it not error in that case.

### DIFF
--- a/examples/common/hf_fsdp.py
+++ b/examples/common/hf_fsdp.py
@@ -6,7 +6,7 @@
 import functools
 from typing import Any, Iterable, List
 
-from transformers import PreTrainedModel
+from transformers import GPTNeoXForCausalLM, PreTrainedModel
 
 # helper functions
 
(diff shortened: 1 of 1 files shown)


samhavens:
IMO this is actually better resolved by removing the function `hf_get_tied_embedding_weights` and replacing it with a call to `causal_base_model.get_input_embeddings()`. `causal_base_model: GPTNeoXModel` [implements this method](https://github.com/huggingface/transformers/blob/v4.26.1/src/transformers/models/gpt_neox/modeling_gpt_neox.py#L430) (I think they all do. My bad not using it initially)

alextrott16:
@samhavens You know this code way better than I do, so I'm happy to go with your suggestion. Do you want to apply those changes? I'm happy to do it instead, but will need a bit more detail from you on what you have in mind. Let me know your preference.

samhavens:
@alextrott16 pushed up what I meant; this works with Neo, Neox, GptJ, gpt2, opt, bloom, whisper's decoder, and codegen (based on inspecting the code for `get_input_embeddings`, did actually attempt to wrap them)

samhavens (approved):
love removing code and covering more cases

PR #148, opened 2023-02-10
pin streaming to <0.2.4

dakinggg:
the distributed init inside of streaming is causing a failure on the tests. And it probably fails in some other not distributed scenarios too.


abhi-mosaic (approved):
LGTM!

PR #150, opened 2023-02-10
Fix YAML referencing to allow better CLI overrides

abhi-mosaic:
Currently we rely on YAML anchors to copy things like `max_seq_len` into many places. 

But since we use OmegaConf to read our YAMLs, we can instead use the `${max_seq_len}` syntax.

This allows CLI to override the anchored values like `data_remote`, `data_local`, `max_seq_len`, etc. that could not be done before. Previously, you would have to override every materialized value like `train_loader.dataset.remote`.

There are a lot of files here, but I'm going to check the high priority ones 1 by 1 to make sure nothing broke. And then assume that YAMLs within a dir (like `yamls/mosaic_gpt/...` all work).


examples/bert/yamls/test/main.yaml
```diff
@@ -3,11 +3,11 @@
 #   `composer main.py yamls/test/main.yaml` to run using the HuggingFace BERT
 #   `composer main.py yamls/test/main.yaml model.name=mosaic_bert` to run using the Mosaic BERT
 
-data_remote: &data_remote ./my-copy-c4 # Follow the instructions in the README to set up ./my-copy-c4
-data_local: &data_local ./my-copy-c4
-tokenizer_name: &tokenizer_name prajjwal1/bert-tiny
-max_seq_len: &max_seq_len 128
-mlm_probability: &mlm_probability 0.15
+data_remote: null # Follow the instructions in the README to set up ./my-copy-c4
```
dakinggg, line 6: did you intentionally change this to null?

examples/llm/icl_eval/my-neo.yaml
```diff
@@ -0,0 +1,44 @@
+# Tokenizer
```
dakinggg, line 1: did you intend to commit this?

examples/llm/yamls/hf_causal_lm/icl-live.yaml
```diff
@@ -0,0 +1,135 @@
+data_local: ./my-copy-c4
```
dakinggg, line 1: also, did you intend to commit?

dakinggg (approved):
yesss thank you. LGTM, assuming you checked a few and the syntax works.

examples/bert/yamls/test/main.yaml
vchiley: yeah if you're not streaming, ie if ds is local, remote should be set to null (see [here](https://github.com/mosaicml/examples/blob/main/examples/llm/yamls/mosaic_gpt/125m.yaml#L1))
we should probably that comment to the line that defines `data_local`

dblalock (commented):
So I'm kind of confused why we still use YAML anchors for the `run_name` in the BERT configs.

- ./examples/bert/yamls/finetuning/glue/mcloud_run.yaml:  base_run_name: *run_name
- ./examples/bert/yamls/finetuning/mcloud_run.yaml:  run_name: *run_name
- ./examples/bert/yamls/main/mcloud_run_a100_40gb.yaml:  run_name: *run_name
- ./examples/bert/yamls/main/mcloud_run_a100_80gb.yaml:  run_name: *run_name

But seems like there are no issues besides those we already flagged.

In more detail, I wrote a script to parse the old and new versions via omegaconf, resolve() them, and then check that they're the same. It correctly found the new `null` that Daniel flagged and said everything else is the same*
so I kinda believe it.

*Except that the BERT yamls can't resolve anything because you need to write out the `parameters:` as a separate yaml file for it to work. (Yet another reason we need to indirect to the other config files).

examples/llm/icl_eval/my-neo.yaml
abhi-mosaic: removing, thank you!

examples/llm/yamls/hf_causal_lm/icl-live.yaml
abhi-mosaic: removing, thank you!

examples/bert/yamls/test/main.yaml
abhi-mosaic: Yes, this was a bug from previous streaming usage where local and remote were set to the same value. Now, we set remote to None

abhi-mosaic:
```
So I'm kind of confused why we still use YAML anchors for the run_name in the BERT configs.
```

I agree, we shouldn't have these. But I think I need to fix the BERT training scripts to read in `COMPOSER_RUN_NAME` correctly to fix. I will look into this, but in a separate PR.

PR #152, opened 2023-02-10
Update Benchmarks README, rename scripts

hanlint:
This PR updates the throughput README for clarity and also renames a few scripts for clarity:
* `run_all_configs.py` -> `submit_benchmarks.py`. Reasoning: the script doesn't always run all configs, as you can subselect certain config files. Also, benchmarks makes it clear what the runs will do.
* `parse_logs.py` -> `collect_results.py`. Collect results is more general, and in the future when we can log metrics directly, we will no longer need to parse the log files. 

Open to other suggestions for script names.


abhi-mosaic (approved):
LGTM, thanks Hanlin. Will add update throughputs today, and more next week as we get more data, for the next release.

PR #155, opened 2023-02-10
Fix wrapping for opt

dakinggg:
Tested on gpt small and opt-2.7b

Screen Shot 2023-02-10 at 1 38 21 PM


examples/common/hf_fsdp.py
```diff
@@ -104,6 +105,9 @@ def prepare_hf_causal_lm_model_for_fsdp(model: PreTrainedModel) -> None:
     HuggingFace for decoder-only LLMs.
     """
     causal_base_model = hf_get_causal_base_model(model)
+    # OPT has an extra layer of wrapping, so special case here
+    if isinstance(causal_base_model, OPTDecoder):
+        model.model._fsdp_wrap = False
```
samhavens, line 110: damn, this was supposed to be caught by
```python
    decoder_attrs = ('transformer', 'model.decoder', 'gpt_neox')
    return findattr(model, decoder_attrs)
```

I wonder why this isn't working

dakinggg: opt has an extra layer, its ComposerHFCausalLM.OPTForCausalLM.OPTModel.OPTDecoder vs ComposerHFCausalLM.GPT2LMHeadModel.GPT2Model. So I'm explicitly setting false for the `OPTModel`, while the existing code is setting false for the `OPTDecoder`.

dakinggg:
Ok @alextrott16 @samhavens , I added a wandb plot. looks like it is training ok.

examples/common/hf_fsdp.py
samhavens: ohh, got it

alextrott16 (approved):
In W&B we trust.

samhavens (approved):
Not the biggest fan of checking one model type explicitly... but I think we are going to change how we do FSDP wrapping soon, so this feels reasonable to me

dakinggg:
Yeah, trying to avoid major changes that might break other things right before release, so we can refactor to be cleaner later. I also added a basic test that test that OPT has the wrap attributes we expect.

examples/llm/tests/test_model.py
```diff
@@ -243,3 +244,21 @@ def test_determinism(attention_type: str, precision):
             loss_2.backward()
             optimizer_1.step()
             optimizer_2.step()
+
+
+def test_opt_wrapping():
+    conf = {
+        'name': 'hf_causal_lm',
+        'pretrained_model_name_or_path': 'facebook/opt-125m',
+        'device': 'cpu',
```
vchiley, line 253: device isn't actually used anywhere in ComposerHFCausalLM

dakinggg: good point haha

dakinggg, line 253:
```suggestion
```

PR #157, opened 2023-02-11
Fix LLM mcloud convert commands

abhi-mosaic:
This PR:
* fixes the data conversion scripts in our LLM `mcloud` YAML files
* fixes a bug with our `mcli-1b-custom.yaml` command to do true 8k context length training


dblalock (approved):
<rubberstamp>

PR #162, opened 2023-02-14
Matthew/convert pile

growlix:
I generalized `convert_c4.py` to work with both The Pile and C4. I renamed it to `convert_dataset.py`, but I also kept an identical version named `convert_c4.py` to facilitate comparison via the diff. The only major changes are adding constants for c4 and The Pile (# samples for different splits, characters per token, etc.), and generalizing the function names and arguments.


bmosaicml (approved):

examples/common/convert_c4.py
```diff
@@ -26,10 +27,12 @@ def parse_args() -> Namespace:
...
     parser = ArgumentParser(
         description=
-        'Convert C4 into MDS format, optionally concatenating and tokenizing')
+        'Convert dataset into MDS format, optionally concatenating and tokenizing')
+    parser.add_argument('--dataset', type=str, required=True)
+    parser.add_argument('--data_subset', type=str, default=None, help='E.g. "all" or "en"')
     parser.add_argument('--splits',
                         nargs='+',
-                        default=['train', 'train_small', 'val'])
+                        default=['train', 'val'])
```
dakinggg, line 35: change this default back maybe?

examples/common/convert_c4.py
```diff
@@ -67,15 +70,120 @@ def parse_args() -> Namespace:
...
+    raw_samples=10000,
+    truncated_samples=10000
+)
+
+
+c4constants = DatasetConstants(
+    chars_per_sample=2163,
+    chars_per_token=4
+)
+pileconstants.splits["train"] = DataSplitConstants(
```
dakinggg, line 148: these ones are probably supposed to be c4?

bmosaicml:
Out of curiosity, could you explain the constants and where they came from, especially these:

chars_per_sample=2163,
 chars_per_token=4

examples/common/convert_c4.py
```diff
@@ -67,15 +70,120 @@ def parse_args() -> Namespace:
...
+class NoConcatDataset(IterableDataset):
     """An IterableDataset that returns text samples for MDSWriter.
 
     Returns dicts of {'text': bytes}
     """
 
     def __init__(self, split: str):
-        self.hf_dataset = hf_datasets.load_dataset(path='c4',
-                                                   name='en',
+        self.hf_dataset = hf_datasets.load_dataset(path='the_pile',
```
dakinggg, line 185: this probably needs to take the same args and not just only work for the pile?

examples/common/convert_c4.py
```diff
@@ -110,21 +218,23 @@ class ConcatTokensC4(IterableDataset):
...
     ):
         self.tokenizer = tokenizer
         os.environ['TOKENIZERS_PARALLELISM'] = 'false'
         self.max_length = max_length
         self.bos_text = bos_text
         self.eos_text = eos_text
         self.should_wrap = not no_wrap
-        self.hf_dataset = hf_datasets.load_dataset(path='c4',
-                                                   name='en',
+        self.hf_dataset = hf_datasets.load_dataset(path=dataset_name,
```
dakinggg, line 236: did you double check that the sharding produces different samples? we had checked c4, but i recall abhi being worried that the sharding might work differently for different datasets.

examples/common/convert_c4.py
```diff
@@ -275,15 +392,17 @@ def generate_samples(
...
     """
-    hf_splits = ('train', 'train', 'validation', 'validation')
-    folder_splits = ('train', 'train_small', 'val', 'val_small')
-    expected_counts = (364868892, 1000000, 364608, 10000)
-    truncate_counts = (None, 100000, None, 10000)
+    try:
+        dataset_constants = CONSTS[args.dataset]
+        iterate_over = dataset_constants
+    except KeyError:
+        warnings.warn(f'Constants for dataset {args.dataset} not found. Will be unable to provide execution time estimates.')
```
dakinggg, line 404: are we confident it works for all hf datasets? or should this actually error instead of warn?

examples/common/convert_c4.py
```diff
@@ -315,19 +444,28 @@ def main(args: Namespace) -> None:
         samples = generate_samples(loader,
                                    truncate_num_samples=truncate_num_samples)
 
-        denominator = truncate_num_samples if truncate_num_samples is not None else _est_progress_denominator(
-            expected_num_samples,
-            mode,
-            args.concat_tokens,
-        )
+        if expected_num_samples:
```
dakinggg, line 447:
```suggestion
        if expected_num_samples is not None:
```

dakinggg (commented):
I left my comments on the `convert_c4.py`.. not sure if that was the right thing to do. also if you remove the old script github should be smart enough to know and just show the diff

growlix:
@bmosaicml 
> Out of curiosity, could you explain the constants and where they came from, especially these:
> 
> chars_per_sample=2163, chars_per_token=4

`chars_per_sample` are computed from the validation sets. I have added this information in comments. `chars_per_token` were obtained by @samhavens using ???.

examples/common/convert_c4.py
growlix: I'm not sure what you mean!
Sharding of the newly-written dataset? Different how? Is this desirable or undesirable behavior?

dakinggg: so in old version of hf datasets, the shards accessed would not be different per worker. so checking this would involve just running a dataloader with multiple workers and making sure each worker's first sample is different.

examples/common/convert_c4.py
dakinggg: Agreed with all the above. I think my worry is that for datasets we haven't tried (e.g. not the pile or c4), this script actually doesn't work for some reason that we haven't discovered yet. So was just wondering how confident we were that it works for all hf datasets.

examples/common/convert_c4.py
growlix: I examined the first 5 batches w/ `num_workers=30` and the content of each batch was different, so I think we're good. Also happy to write up a more comprehensive test if you'd advise that.

dakinggg: That is good enough for me for now. We're missing a lot of test coverage, and I think this is ok.

examples/common/convert_c4.py
growlix: I think we're not confident. Should we have it error out if a dataset is selected that's not in `CONSTANTS`?

dakinggg: I think lets error out if its not pile/c4 for now. We can easily add/test more, and a user can easily remove that error and use whatever they want at their own risk. Maybe have the error say something like "we've only tests on c4/pile. it may extend to other hf datasets. please open an issue if one interests you!"

examples/llm/tests/test_c4_data_prep_script.py
```diff
@@ -5,7 +5,7 @@
 import shutil
 from argparse import Namespace
 
-from examples.common.convert_c4 import main
+from examples.common.convert_dataset import main
```
dakinggg, line 8: need to change the name in the command line version of this test too

dakinggg (approved):
LGTM apart from remaining comments

growlix:
Generalizes `convert_c4.py` to work with both The Pile and C4. `convert_c4.py` is renamed to `convert_dataset.py`, Major changes are: adding constants for c4 and The Pile (# samples for different splits, characters per token, etc.), and generalizing the function names and arguments.

PR #163, opened 2023-02-14
Fix `key_padding_mask` when using `attn_impl='flash'`

abhi-mosaic:
When we merged #128 , we removed the codepath by which we always force a padding token `pad_token = eos_token` for tokenizers like GPT2. Since this can lead to silent errors. We also started pretokenizing our data and passing the raw torch tensor to the HF collate fn, rather than a batch created by a HF tokenizer (with `attention_mask` optionally inside).

These changes both lead to our text dataloader batches not producing `attention_mask` anymore with the GPT2 tokenizer, it just produces `None`, since there is no padding possible anymore.

This meant that our `MosaicGPT` class was now receiving `key_padding_mask=None` in all situations, including `attn_impl='flash'`. We had not tested this before because it had never showed up.

Upon initial testing of `attn_impl=flash` up to 3B models, nothing seemed broken, training throughput was fine, and even slightly higher for some models.

Last night, while testing 7B models with the settings in our `throughput/` tables, we noticed that the MFU was significantly worse, and there was a lot of "thrashing" in the memory allocator. 

The `FlashMHA` layer from HazyResearch has a separate codepath when `key_padding_mask=None`: https://github.com/HazyResearch/flash-attention/blob/2dc2a195890d323f6f9e1b74e4667099e6144f79/flash_attn/flash_attention.py#L42

For reasons unknown, this codepath is hurting performance for the larger models. Avoiding this codepath by sending a `key_padding_mask` of all 1s appears to fix the problem.

### DIFF
--- a/examples/llm/src/mosaic_gpt.py
+++ b/examples/llm/src/mosaic_gpt.py
@@ -423,7 +423,7 @@ class MosaicGPT(nn.Module):
                                     dtype=x.dtype)
         if self.cfg.attn_impl == 'flash' and key_padding_mask is None:
             mod_key_padding_mask = torch.ones_like(input_ids, dtype=torch.bool)
-        if self.cfg.attn_impl == 'triton':
+        elif self.cfg.attn_impl == 'triton':
             mod_key_padding_mask = None
         else:
             mod_key_padding_mask = key_padding_mask
--- a/examples/llm/src/mosaic_gpt.py
+++ b/examples/llm/src/mosaic_gpt.py
@@ -421,9 +421,9 @@ class MosaicGPT(nn.Module):
                                     seq_len=S,
                                     key_padding_mask=key_padding_mask,
                                     dtype=x.dtype)
-        if cfg.attn_impl == 'flash':
+        if self.cfg.attn_impl == 'flash':
             key_padding_mask = torch.ones_like(input_ids, dtype=torch.bool)
-        if cfg.attn_impl == 'triton':
+        if self.cfg.attn_impl == 'triton':
             key_padding_mask = None
         for block in self.transformer.blocks:  # type: ignore
             x = block(x, key_padding_mask, attn_mask)
--- a/examples/llm/src/mosaic_gpt.py
+++ b/examples/llm/src/mosaic_gpt.py
@@ -422,6 +422,9 @@ class MosaicGPT(nn.Module):
                                     key_padding_mask=key_padding_mask,
                                     dtype=x.dtype)
         if self.cfg.attn_impl == 'flash' and key_padding_mask is None:
+            # HazyResearch FlashMHA appears to use more memory when `key_padding_mask=None`
+            # in certain settings like MosaicGPT-7B. So we always provide a tensor.
+            # See https://github.com/mosaicml/examples/pull/163 for more details.
             mod_key_padding_mask = torch.ones_like(input_ids, dtype=torch.bool)
         elif self.cfg.attn_impl == 'triton':
             mod_key_padding_mask = None
--- a/examples/llm/src/mosaic_gpt.py
+++ b/examples/llm/src/mosaic_gpt.py
@@ -421,12 +421,14 @@ class MosaicGPT(nn.Module):
                                     seq_len=S,
                                     key_padding_mask=key_padding_mask,
                                     dtype=x.dtype)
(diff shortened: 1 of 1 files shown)


examples/llm/src/mosaic_gpt.py
```diff
@@ -421,10 +421,12 @@ def forward(self,
...
                                     dtype=x.dtype)
+        if self.cfg.attn_impl == 'flash':
+            key_padding_mask = torch.ones_like(input_ids, dtype=torch.bool)
+        if self.cfg.attn_impl == 'triton':
+            key_padding_mask = None
         for block in self.transformer.blocks:  # type: ignore
-            x = block(
-                x, None if self.cfg.attn_impl == 'triton' else key_padding_mask,
-                attn_mask)
+            x = block(x, key_padding_mask, attn_mask)
```
dskhudia, line 429: Should the triton path still receive None? Also by moving to the pre-tokenized datasets, we no longer support tokenization on the fly with MosaicGPT models right? as for tokenization on the fly setting key_padding_mask to 1 is not right.

examples/llm/src/mosaic_gpt.py
```diff
@@ -421,10 +421,12 @@ def forward(self,
                                     seq_len=S,
                                     key_padding_mask=key_padding_mask,
                                     dtype=x.dtype)
+        if self.cfg.attn_impl == 'flash':
+            key_padding_mask = torch.ones_like(input_ids, dtype=torch.bool)
+        if self.cfg.attn_impl == 'triton':
```
dakinggg, line 426:
```suggestion
        elif self.cfg.attn_impl == 'triton':
```
for clarity that these are mutually exclusive code paths

examples/llm/src/mosaic_gpt.py
```diff
@@ -421,10 +421,12 @@ def forward(self,
                                     seq_len=S,
                                     key_padding_mask=key_padding_mask,
                                     dtype=x.dtype)
+        if self.cfg.attn_impl == 'flash':
+            key_padding_mask = torch.ones_like(input_ids, dtype=torch.bool)
```
dakinggg, line 425: doesn't this need to actually take into account padding? tokenization on the fly is still supported, as long as you use a valid tokenizer.

examples/llm/src/mosaic_gpt.py
abhi-mosaic: The `triton` path with `None` is just copying what was there before, I am not sure of its correctness. @vchiley thoughts here?

For on the fly padding and tokenization you are right, that is a bug on my part. Will modify the check appropriately.

examples/llm/src/mosaic_gpt.py
dakinggg: now it needs the elif, otherwise will ignore the setting in the flash conditional and hit the else

examples/llm/src/mosaic_gpt.py
vchiley: triton doesn't support key_padding_mask, if key_padding_mask shows invalid tokens, you need to handle it as part of the attn_bias (at least it wasn't supported last time i looked at it; with time that might change; i there is an [issue in HazyR](https://github.com/HazyResearch/flash-attention/issues/104) asking for it to be supported)

dakinggg (commented):
This looks good to me now, but I'm going to leave approval because I'd like another pair of eyes since its a late breaking fix. cc @dskhudia or @vchiley for another look

examples/llm/src/mosaic_gpt.py
```diff
@@ -421,10 +421,14 @@ def forward(self,
                                     seq_len=S,
                                     key_padding_mask=key_padding_mask,
                                     dtype=x.dtype)
+        if self.cfg.attn_impl == 'flash' and key_padding_mask is None:
+            mod_key_padding_mask = torch.ones_like(input_ids, dtype=torch.bool)
```
dskhudia, line 425: A small comment here on why this is necessary (may be with PR link) will be useful for anyone reading this in future.

dskhudia (approved):

vchiley (approved):
if throughput looks good, then this lgtm

abhi-mosaic:
Verified with a 7B model on 80GB and 40GB cards, and throughput looks good. Adding a comment with link to this PR and then merging.

PR #164, opened 2023-02-14
Add sweep for throughputs

abhi-mosaic:
* fix bug where `git_branch` default prevents `git_commit` from being used
* Add `sweep.sh` script to easily reproduce throughput tables with hardcoded settings.


dakinggg (approved):
LGTM

PR #169, opened 2023-02-16
[XS] Fix README `cd`

jbloxham:
The previous command made it look like `examples/examples/llm` was the path


dakinggg:
`examples/examples/llm` is the path :) the first `examples` is the repo, the second `examples` is the package. I tried to make this clear with the inline comments, but clearly did not succeed. Any suggestion for how to make it more clear?

PR #174, opened 2023-02-16
Query Key LayerNorm for Flash Attention

bcui19:
Adds in query key layernorms for flash attention:

The associated JIRA:
https://mosaicml.atlassian.net/jira/software/c/projects/RESEARCH/boards/24?modal=detail&selectedIssue=RESEARCH-444&assignee=63127d143778a7aadf1901c7


examples/llm/src/mosaic_gpt.py
```diff
@@ -86,26 +87,74 @@ class FlashCausalAttention(nn.Module):
...
+        if self.attn_qk_ln:
+            qkv = self.W_qkv(x)
+
+            # Applying layernorm to qk
+            dtype = qkv.dtype
+            q, k, v = qkv.split(self.d_model, dim=-1)
+            q = self.q_ln(q)
+            k = self.k_ln(k)
+            qkv = torch.cat([q, k, v], dim=-1)
+            qkv = qkv.to(dtype)
```
abhi-mosaic, line 138: Whats the purpose of the `dtype` stuff here? Shouldn't autocast make sure that the activations are always 16-bit as they are being passed around?

examples/llm/src/mosaic_gpt.py
```diff
@@ -86,26 +87,74 @@ class FlashCausalAttention(nn.Module):
...
+                                              device=device)
+            self.out_proj = nn.Linear(self.d_model,
+                                      self.d_model,
+                                      bias=True,
+                                      device=device)
+
+            self.out_proj._is_residual = True  # type: ignore
+
+            self.q_ln = nn.LayerNorm(self.d_model, device=device)
+            self.k_ln = nn.LayerNorm(self.d_model, device=device)
```
abhi-mosaic, line 114: Just want to double check here that 2 LayerNorms are required and not just a shared one for both activations.

abhi-mosaic (approved):
Approving as it looks reasonable to me. Would love to see a side-by-side run on **this branch** with say, a 125M model with `cfg.model.attn_qk_ln = [True | False]`

examples/llm/src/mosaic_gpt.py
bcui19: I think layernorms upcast everything to fp32. As a result, we need to downcast back to the original datatype (in particular flash doesn't like fp32)

examples/llm/src/mosaic_gpt.py
bcui19: Yeah how it's formulated is separately

examples/llm/src/mosaic_gpt.py
abhi-mosaic: Hmm but wouldn't that mean all our networks with autocast have FP32 activations running around? That doesn't sound right...

vchiley, line 138:
```suggestion
            # Applying layernorm to qk
            dtype = qkv.dtype
            q, k, v = qkv.split(self.d_model, dim=-1)
            q = self.q_ln(q).to(dtype)
            k = self.k_ln(k).to(dtype)
            qkv = torch.cat([q, k, v], dim=-1)
```

examples/llm/src/mosaic_gpt.py
```diff
@@ -86,26 +87,74 @@ class FlashCausalAttention(nn.Module):
...
-            embed_dim=cfg.d_model,
-            num_heads=cfg.n_heads,
-            attention_dropout=cfg.attn_pdrop,
-            bias=True,
-            batch_first=True,
-            causal=True,
-            device=device,
-        )
-        self.mhsa.out_proj._is_residual = True
+        self.attn_qk_ln = cfg.attn_qk_ln
```
vchiley, line 95:
```suggestion
        self.attn_qk_ln = cfg.get('attn_qk_ln')
```
if we do it this way, we don't need to add it to the config unless we're using it.

examples/llm/src/mosaic_gpt.py
```diff
@@ -264,6 +313,11 @@ def __init__(self, cfg: DictConfig):
         else:
             raise ValueError(f'Unknown attn_impl={cfg.attn_impl}')
 
+        if cfg.attn_qk_ln and cfg.attn_impl != 'flash':
```
vchiley, line 316:
```suggestion
        if cfg.get('attn_qk_ln') and cfg.attn_impl != 'flash':
```

examples/llm/yamls/mosaic_gpt/125m.yaml
```diff
@@ -23,6 +23,7 @@ model:
   resid_pdrop: 0.0
   emb_pdrop: 0.0
   attn_impl: flash
+  attn_qk_ln: false
```
vchiley, line 26:
```suggestion

```
remove from all configs; we don't need it unless we need it.

vchiley (approved):
few minor comments.

PR #175, opened 2023-02-16
Upgrade to torch 13

mvpatel2000:
As title. Seems to have no throughput hit and makes xformers easier

### DIFF
--- a/examples/stable_diffusion/requirements.txt
+++ b/examples/stable_diffusion/requirements.txt
@@ -1,6 +1,6 @@
 # --------- pytorch --------- #
-torch>=1.12.0
-torchvision>=0.11.0
+torch>=1.13.0
+torchvision>=0.12.0
 mosaicml[all]>=0.12.1,<0.13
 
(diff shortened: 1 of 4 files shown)


examples/stable_diffusion/requirements.txt
```diff
@@ -10,6 +10,5 @@ omegaconf
 diffusers[torch]==0.11.1
 transformers[torch]==4.25.1
 
-# -------- optional -------- #
-# ninja # Faster xformers install
-# git+https://github.com/facebookresearch/xformers.git@3df785ce54114630155621e2be1c2fa5037efa27#egg=xformers #xformers
+# -------- xformers -------- #
```
abhi-mosaic, line 13: is it a good idea to update lines L2-L3 above so that it forces torch 1.13?

mvpatel2000:
I've tested on multinode and am unable to find 1.12 vs. 1.13 discrepencies...

A-Jacobson (approved):

PR #177, opened 2023-02-17
Use round-trip time for `SpeedMonitor`

abhi-mosaic:
In situations when heavy callbacks are used on `batch_end`, such as `OptimizerMonitor`, the overhead was not being reflected in the `SpeedMonitor`'s metrics.

This refactor cuts out some of the state variables and makes the logic a bit simpler.

Note that because we only count the timestamps at `batch_end`, if you use a `window_size=N`, it will not log the througput metrics until batch `N+1`

### DIFF
--- a/examples/common/speed_monitor_w_mfu.py
+++ b/examples/common/speed_monitor_w_mfu.py
@@ -171,7 +171,7 @@ class SpeedMonitorMFU(Callback):
         # Log the throughput
         if len(self.history_wct) == self.history_wct.maxlen:
             world_size = dist.get_world_size()
-            elapsed_batches = len(self.history_samples)
+            elapsed_batches = len(self.history_samples) - 1
             elapsed_samples = int(self.history_samples[-1]) - int(
                 self.history_samples[0])
             elapsed_wct = self.history_wct[-1] - self.history_wct[0]
--- a/examples/common/speed_monitor_w_mfu.py
+++ b/examples/common/speed_monitor_w_mfu.py
@@ -140,11 +140,8 @@ class SpeedMonitorMFU(Callback):
                  window_size: int = 100,
                  gpu_flops_available: Optional[Union[float, int]] = None):
         # Track the batch num samples and wct to compute throughput over a window of batches
-        self.batch_start_num_samples = 0
-        self.batch_start_wct = 0.0
-        self.batch_wct_buffer: Deque[float] = deque(maxlen=window_size)
-        self.batch_num_samples_buffer: Deque[int] = deque(maxlen=window_size)
-        self.window_size = window_size
+        self.history_samples: Deque[float] = deque(maxlen=window_size + 1)
+        self.history_wct: Deque[int] = deque(maxlen=window_size + 1)
 
         self.set_gpu_flops_available = False
         self.gpu_flops_available = gpu_flops_available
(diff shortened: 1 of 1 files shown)


abhi-mosaic:
Did a run with checkpointing every 5ba, then resuming from batch 10, to make sure nothing broke in the state resumption.
Screenshot 2023-02-16 at 6 00 15 PM

mvpatel2000 (commented):
This is a lot cleaner and simpler, but I think we should keep the code as is and just do a simple patch (something like https://github.com/mosaicml/composer/pull/1977). Unfortunately, this PR breaks backwards compatibility for checkpoints, which is bad :( and also means SpeedMonitor used from examples can't be replaced with Composer SpeedMonitor for resumption (which also might cause problems)

mvpatel2000 (changes requested):
Holding based on above concerns -- discussing offline

abhi-mosaic:
Thanks for the context @mvpatel2000 , I see the need for backwards compatibility.

I just added a commit that removes nearly all state for resumption. So now, older checkpoints are still valid (we just won't use some keys). I don't think we are commiting to new checkpoints loaded in old code, because that would still break.

What happens without state? Well if you resume on batch 20 and you have a `window_size=5`, you will have to wait till batch 26 for the throughput to appear. Previously it would appear on batch 21 right away, but it would carry over history (like any slow batches) from the previous run. I'll post a screenshot below, but I think it actually looks a bit better now.

P.S. the last bit of state remaining `total_eval_wct` kinda feels like something that Composer.state.timestamp should have... it would certainly make this callback safer if it was totally stateless and just did math on timestamp objects.

abhi-mosaic:
Here's what resumption looks like now:
Screenshot 2023-02-16 at 8 07 30 PM

mvpatel2000:
> What happens without state? Well if you resume on batch 20 and you have a `window_size=2`, you will have to wait till batch 23 for the throughput to appear. Previously it would appear on batch 21 right away, but it would carry over history (like any slow batches) from the previous run. I'll post a screenshot below, but I think it actually looks a bit better now.

This refactor is a lot cleaner. However, I don't really like losing functionality for code cleanliness -- it seems much better to just have a smooth resumption and a continuous line. Is there a good reason you think this change (removing things from state dict) is worth losing functionality? 

> P.S. the last bit of state remaining `total_eval_wct` kinda feels like something that Composer.state.timestamp should have... it would certainly make this callback safer if it was totally stateless and just did math on timestamp objects.

Yea... the timestamp design of only running in eval isn't very helpful during training... Will think about this

abhi-mosaic:
I think of the SpeedMonitor's job as logging the speed of the current run, which feels like it should be true to the current run only, not holding over the slow history of previous runs.

With the old behavior, a resumption from 8 GPUs to 128 GPUs would have a few batches at the old speed, like this one:
image

The starts of the green and red lines are artificially reduced for a block of `window_size` time (because of the loaded state), but it's not accurate, those runs were fast right away. 

I feel like its a style thing ? But one of them has the possibility of misinterpretation whereas the other just doesn't report anything until the buffer is filled?

abhi-mosaic:
If the goal is continuity, I think we could allow the window to have warmup.. like if we report noisier estimates at `window_size=1, window_size=2, ... window_size=window_size`, and then roll from there?

mvpatel2000 (approved):
LGTM! Thanks for hopping on this so quickly. I can handle upstreaming for v13 release

abhi-mosaic:
Ok! I will merge this to get in before the 7B run today, but if you want to try and play around with ways of getting better continuity, totally open to that. Thanks for all the push to make things backwards compat!

PR #178, opened 2023-02-17
Add support for using TransformerEngine

dskhudia:
Use TransformerEngine for either 1) linear layers only 2) Whole te.TransformerLayer to replace GPTBlock.

### DIFF
--- a/examples/llm/main.py
+++ b/examples/llm/main.py
@@ -18,6 +18,11 @@ from examples.common.config_utils import log_config, update_batch_size_info
 from examples.llm.src.model_registry import COMPOSER_MODEL_REGISTRY
 from examples.llm.src.tokenizer import TOKENIZER_REGISTRY
 
+try:
+    import transformer_engine.pytorch as te
+    te_installed = True
+except ImportError:
+    te_installed = False
 
 def build_composer_model(cfg):
     warnings.filterwarnings(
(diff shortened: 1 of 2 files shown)


abhi-mosaic:
Hi @dskhudia, can you comment on how necessary it is to use the `TransformerLayer` for performance / convergence?

For context, we have been making many changes to `GPTBlock` (and expect to make more variations), supporting different attention / arch / normalization strategies, and I'm not sure if we would end up using `TransformerLayer` much. The `te.Linear` approach feels more modular and we could use it with all the variations we end up building.

If it's still useful for testing purposes to have `TransformerLayer`, then I can think about a way to make the model construction more modular like `cfg.block_type=...`

examples/llm/main.py
```diff
@@ -64,6 +69,23 @@ def main(cfg):
             "Reverting to `cfg.model.init_device='cpu'`.")
         cfg.model.init_device = 'cpu'
 
+    te_config = cfg.get('te_config', None)
+    te_config = om.to_container(te_config, resolve=True) if te_config else None
+    if not te_installed and te_config:
+        warnings.warn(
```
hanlint, line 75: Do we need to check whether it's an H100 and alert appropriately?

examples/llm/main.py
```diff
@@ -64,6 +69,23 @@ def main(cfg):
...
+                "Transformer Engine is not installed but te_config is given! "
+                "Please install transformer engine with "
+                "pip install --upgrade git+https://github.com/NVIDIA/TransformerEngine.git@stable")
+    if not te_installed and not te_config['linear_only']:
+        print('use attn_impl == Triton to take care of key_padding_mask with TransformerEngine...')
```
hanlint, line 80: log instead of print. Message should also be more explicit: "Overriding attn_impl to use Triton to be compatible with Transformer Engine."

examples/llm/main.py
```diff
@@ -64,6 +69,23 @@ def main(cfg):
...
+        cfg.attn_impl = 'triton'
+
+    if te_installed and te_config and init_device == 'meta':
+        warnings.warn(
+            "Using `cfg.model.init_device='meta'` doesn't work with transformer engine! "
```
hanlint, line 85:
```suggestion
            "Using `cfg.model.init_device='meta'` not compatible with Transformer Engine."
```

examples/llm/main.py
```diff
@@ -64,6 +69,23 @@ def main(cfg):
...
+            "Reverting to `cfg.model.init_device='cpu'`.")
+        cfg.model.init_device = 'cpu'
```
hanlint, line 87: open question: should we be overriding settings vs. throwing an error? e.g. forcing the init device out of meta can have other implications, and maybe forcing the user to switch it might be a better experience?

examples/llm/src/mosaic_gpt.py
```diff
@@ -205,13 +210,21 @@ class GPTMLP(nn.Module):
 
     def __init__(self, cfg: DictConfig, device: Optional[str] = None):
         super().__init__()
-        self.mlp_up = nn.Linear(cfg.d_model,
-                                cfg.mlp_ratio * cfg.d_model,
-                                device=device)
+        if te_installed and cfg.get('te_config.linear_only'):
```
hanlint, line 213: should the requisite warnings (e.g. if te_config provided but te not installed) but put here as well, if users are creating their own models and not using `main.py` ?

SinanAkkoyun:
Hello!
How does one use this with inference of the model? We have a great interest in that.

Best Regards

dakinggg:
Closing, because llm has moved to llm-foundry.

PR #179, opened 2023-02-17
Add/refactor HF models in LLM

alextrott16:
This slightly refactors existing model code and extends HF support to 3 types of LLM models: Causal LMs, Prefix LMs, and T5 models. It unifies the construction of each in a way that makes more usage of Composer's `HuggingFaceModel`.

This PR is part of a set of code changes that will integrate UL2 training and other seq-to-seq fine-tuning tools into LLM.


examples/llm/src/models/utils/meta_init_context.py
```diff
@@ -0,0 +1,128 @@
+# Copyright 2022 MosaicML Examples authors
+# SPDX-License-Identifier: Apache-2.0
+
+# Copyright 2022 The HuggingFace Team. All rights reserved.
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
```
abhi-mosaic, line 10: @alextrott16 could you check with @hanlint what we have to do re. licenses here? If this gets complicated, we can rewrite the meta init function ourselves or try to use something like `torchdistx` with `deferred_init()`

hanlint: LGTM

abhi-mosaic:
@alextrott16 could you provide me an examples command to test training, say, a T5 model? I just want to try to go through the workflow.

I would also be in favor of merging this soon (#182 is also waiting) and patching it as we use it and find issues

dakinggg (approved):
I think this looks good to me. Its hard to verify alll the code, so proof is in the pudding. would be nice to see the start of a normal gpt neo run (which we know works prior to this pr), and the start of a t5 run (which is becoming supported in this pr), and maybe the start of a prefix lm run?

alextrott16:
@abhi-mosaic @dakinggg This PR is not supposed to be standalone, just the first of a few that will add more full support for T5, PrefixLM, UL2, and seq2seq fine-tuning. It does not yet enable T5 training because there is still not code that supports that style of dataloading. I have at least added some test coverage to check that the T5 models and PrefixLM models handle a forward/backward pass.
I have also locally verified that Causal LM and Prefix LM training work. And I've successfully done ICL evaluation on both in an interactive session.

dakinggg:
Great, that is good enough for me, thanks!

PR #181, opened 2023-02-21
allow users to set custom groups for glue finetuning on wandb

codestar12:
give users the ability to set custom groups for wandb logging


dakinggg (approved):
LGTM, thanks

PR #185, opened 2023-02-22
Moving attention into a separate layers file

bcui19:
As title. Trying to clean up the file structure of GPT.


vchiley (changes requested):
- add `__init__.py` file to `layers/` and `models/`

- mv `GPTBlock` out of `MosaicGPT.py` and into `layers/`

vchiley (approved):

PR #186, opened 2023-02-22
added lion

bmosaicml:
Testing with this run: gpt-125m-lion-q6aul5

### DIFF
--- a/examples/common/builders.py
+++ b/examples/common/builders.py
@@ -4,7 +4,7 @@
 import os
 
 from composer import algorithms
-from composer.callbacks import LRMonitor, MemoryMonitor, OptimizerMonitor
+from composer.callbacks import LRMonitor, MemoryMonitor, OptimizerMonitor, SpeedMonitor
 from composer.core import Evaluator
 from composer.datasets.in_context_learning_evaluation import \
     get_icl_task_dataloader
@@ -24,7 +24,7 @@ def build_callback(name, kwargs):
     elif name == 'memory_monitor':
         return MemoryMonitor()
     elif name == 'speed_monitor':
-        return SpeedMonitorMFU(window_size=kwargs.get('window_size', 1),
+        return SpeedMonitor(window_size=kwargs.get('window_size', 1),
                                gpu_flops_available=kwargs.get(
                                    'gpu_flops_available', None))
     elif name == 'optimizer_monitor':
--- a/examples/common/builders.py
+++ b/examples/common/builders.py
@@ -130,7 +130,8 @@ def build_icl_evaluators(cfg, tokenizer):
                 num_fewshot=num_fewshot,
                 prompt_string=icl_cfg.prompt_string,
                 example_delimiter=icl_cfg.example_delimiter,
-                continuation_delimiter=icl_cfg.continuation_delimiter)
+                continuation_delimiter=icl_cfg.continuation_delimiter,
+                destination_path=icl_config.destination_path)
             logger_keys.extend([f'metrics/{label}/{m}' for m in metric_names])
             evaluators.append(
                 Evaluator(label=label,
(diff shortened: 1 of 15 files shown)


examples/common/optim/lion.py
```diff
@@ -0,0 +1,134 @@
...
+        lr: float = 1e-4,
+        betas: Tuple[float, float] = (0.9, 0.99),
+        weight_decay: float = 0.0,
+        use_triton: bool = False
+    ):
+        assert lr > 0.
+        assert all([0. <= beta <= 1. for beta in betas])
+        if weight_decay >= 1e-3:
+            log.warning(
+                f'You are using a high value of `weight_decay={weight_decay}` for the `DecoupledAdamW` optimizer. Are you sure you want to do this? '
```
abhi-mosaic, line 63:
```suggestion
                f'You are using a high value of `weight_decay={weight_decay}` for the `DecoupledLionW` optimizer. Are you sure you want to do this? '
```

abhi-mosaic (commented):
Couple high level comments:

* lets remove Triton for now, we don't use any fused optimizers in our other work and I'd rather not add complexity. I don't expect this to change runtime much as we use large batch sizes and ZERO-2/3
* can we rename `update_fn` to `lionw` to match the style of our other optimizers and torch? https://github.com/mosaicml/composer/blob/c9b7524c941d10196432d0dc145d0fcdc3376cbf/composer/optim/decoupled_weight_decay.py#L228
* can we triple check the implementation against the paper (and not just the lion-pytorch codebase :)

examples/common/builders.py
```diff
@@ -63,6 +63,10 @@ def build_optimizer(cfg, model):
                               betas=cfg.betas,
                               eps=cfg.eps,
                               weight_decay=cfg.weight_decay)
+    elif cfg.name == 'decoupled_lionw':
+        return DecoupledLionW(model.parameters(),
+                    lr=cfg.lr,
```
abhi-mosaic, line 68: need to add other kwargs here

examples/common/optim/lion.py
```diff
@@ -0,0 +1,129 @@
...
+
+    @staticmethod
+    def lionw(p, grad, exp_avg, lr, initial_lr, wd, beta1, beta2) -> None:
+        # stepweight decay
+        if wd != 0:
+            decay_factor = (lr / initial_lr) if initial_lr else 1.0
+            p.data.mul_(1 - decay_factor * wd)
+
+        # update is interpolation between gradient and momentum
+        update = exp_avg.clone().lerp_(grad, 1 - beta1).sign_()
```
abhi-mosaic, line 72:
```suggestion
        update = exp_avg.lerp(grad, 1 - beta1).sign_()
```

I think this should create a new tensor without a need for a .clone()?

abhi-mosaic (approved):
Made one suggestion and approving to unblock.

@bmosaicml could you post a screenshot of two runs, one with DecoupledAdamW and one with DecoupledLionW to confirm that stuff is working? Ideally also show the `reserved_memory` which should be lower for the DecoupledLionW run.

examples/common/builders.py
```diff
@@ -63,6 +63,12 @@ def build_optimizer(cfg, model):
                               betas=cfg.betas,
                               eps=cfg.eps,
                               weight_decay=cfg.weight_decay)
+    elif cfg.name == 'decoupled_lionw':
+        return DecoupledLionW(model.parameters(),
+                    lr=cfg.get('lr', 1e-4),
+                    betas=cfg.get('betas', (0.9, 0.99)),
+                    weight_decay=cfg.get('weight_decay', 0.0)
```
abhi-mosaic, line 70: I would prefer not to allow any defaults for optimizer hparams... I want our YAMLs to be forced to specify each hparam so it's easily readable without having to dig into the github commit for a particular run.

```suggestion
                    lr=cfg.lr,
                    betas=cfg.betas,
                    weight_decay=cfg.weight_decay
```

PR #187, opened 2023-02-23
refactor file structure and attn mask gen fn

vchiley:
- moving layers into models folder
- mv attn mask gen fn
- updt test

all tests pass


abhi-mosaic (approved):
This looks good to me, mostly moving code.. relying on tests to make sure nothing broke. 

Thank you @vchiley !!

PR #189, opened 2023-02-23
UL2 support

alextrott16:
- Adds a mixture of denoisers collator and dataloader constructor: `src/data/denoising.py`.
- Adds test coverage to ensure correct dataloading with and without pre-tokenization.
- Updates `main.py` to provide a `"text_denoising"` dataloader option, and adds a dataloader/model compatibility check.


alextrott16:
Note about all the commits: I was developing this code on branch `alex/ul2`. I made this branch to stage some of that code and must've accidentally brought all development code with it. Almost all the commits are from work on the `alex/ul2` branch, which was building up code in a side folder `examples/examples/ul2`. I didn't mean to include that code, so I committed a change that deletes that side branch, rendering the first 30 or so commits meaningless.

The first meaningful commit is `Add denoising dataloader`.

examples/llm/main.py
```diff
@@ -29,7 +53,16 @@ def build_composer_model(cfg):
         raise ValueError(f'Not sure how to build model with name={cfg.name}')
 
 
+def build_dataloader(cfg, device_batch_size):
+    if cfg.name == 'text_denoising':
+        return build_text_denoising_dataloader(cfg, device_batch_size)
+    return _build_dataloader(cfg, device_batch_size)
```
abhi-mosaic, line 59: oh this is kinda obscure naming... should we augment the `_build_dataloader` inside `builders.py` so that it can switch on `cfg.name`?

examples/llm/src/data/denoising.py
```diff
@@ -0,0 +1,705 @@
...
+
+        # Truncate portions of the inputs that are purely padding
+        # (up to a multiple of 8)
+        multiple_of = 8
+        n_examples_per_length = batch['attention_mask'].sum(0)
+        keep_tokens = torch.sum(n_examples_per_length > 0)
+        keep_tokens = int(multiple_of * torch.ceil(keep_tokens / multiple_of))
+
+        # Note: EncDec formatting will always produce a right-padded batch
+        if self.tokenizer.padding_side == 'left' and self.decoder_only_format:
```
abhi-mosaic, line 226: is there any case where we want to support left-padding? I think it presents a bit of a challenge for our causal LM with certain attention implementations, but then again I guess this denoiser will never be used by causal LMs...

examples/llm/main.py
alextrott16: I'm fine with this, but it will add a very non-common (i.e. specific to LLM) dataloader in the common dataloader builder.

abhi-mosaic (approved):
Approving to allow forward progress on UL2. Will fully admit that I can't verify all the nuances of all the denoising strategies. 

the things I'm most concerned about:
* are the different denoising strategies implemented correctly? Would be best to add some test cases but can be a followup PR
* give the variable seqlens produced, are there any concerns about the train microbatch size needing to change give a different value of `mask_ratio`? Like if i use a really big ratio, will I produce longer sequences, need more memory, and have to use smaller microbatches? Just want to check my intuition.

examples/llm/src/data/denoising.py
alextrott16: My intuition was that the collator should not override the padding side choice (except in the weird case where someone tries to use a left-padded-by-default tokenizer, for example, with EncDec formatting).

You're correct that this collator shouldn't be used with causal LMs, but if there's a causal LM that can't support a given padding side, its prefix LM version will have the same limitation. The good news is that I think HF has been careful to match the default padding side of a tokenizer with the expected padding side of any model implementation. I think that, by supporting both padding sides, this collator helps to take advantage of that.

examples/llm/main.py
abhi-mosaic: Oh i see what you mean... hmm well if you think that the denoising dataloader won't be used elsewhere only here, perhaps we can just do:

```
from examples.common.text_data import build_text_dataloader
...
 if cfg.name == 'text_denoising':
        return build_text_denoising_dataloader(cfg, device_batch_size)
elif cfg.name == 'text':
    return return build_text_dataloader(cfg, device_batch_size)
else:
  raise ValueError(...)
```

just to make it easier to grok what's going on and click through to the relevant dataloader code

alextrott16:
@abhi-mosaic Thanks for the feedback and quick turnaround! I'll address your two concerns:
- I'll add a TODO for a more thorough test of the noising, but I feel very confident that it behaves correctly based on lots and lots of qualitative checks.
- The code is very careful to respect the `max_seq_len` argument. It will pre-truncate the token sequences before doing any span masking, adding prefix tokens, or adding an EOS token (if needed). That pre-truncation ensures that `max_seq_len` will be respected. With that said, different denoiser mixtures will result in differences in sequence lengths because excess padding is trimmed where possible. In certain cases, you might be able to handle a larger micro-batch size for one mixture versus another. But, in those cases, you'll still enjoy higher throughput regardless of if you tune your batch size around the mixture statistics. In sum, though, I think it work well to treat `max_seq_len` as the expected sequence length and to plan micro-batch sizes based on that (in the same way you would when training a causal LM).

PR #190, opened 2023-02-24
adding dist init to main

vchiley:
If dist is not init in main, trainer will init it
If model init needs dist (ie MoE models), then dist must be init in model init

This PR makes sure this isn't an issue going forward.

### DIFF
--- a/examples/llm/main.py
+++ b/examples/llm/main.py
@@ -7,7 +7,7 @@ import warnings
 
 from composer import Trainer
 from composer.core import Evaluator
-from composer.utils import reproducibility
+from composer.utils import dist, get_device, reproducibility
 from omegaconf import OmegaConf as om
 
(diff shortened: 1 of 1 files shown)


examples/llm/main.py
```diff
@@ -39,6 +39,7 @@ def main(cfg):
     )
 
     reproducibility.seed_all(cfg.seed)
+    dist.initialize_dist(get_device(None))
```
dblalock, line 42: can we not do
```python
    dist.initialize_dist(get_device())
```
or just 
```
dist.initialize_dist()
```
? I assume not since you wrote it this way; if not, we should open a Composer usability issue

dblalock (approved):
LGTM

examples/llm/main.py
vchiley: I will open a Composer usability issue

PR #192, opened 2023-02-25
Modifications to allow uploading Mosaic Bert to the hub

dakinggg:
The modifications are:
- add a `BertConfig` class with our specific params. It inherits from `transformers.BertConfig`
- change our `from_pretrained` to be called `from_composer`, so that the base class `from_pretrained` is still a thing

The hub model can be tested via `transformers.AutoModelForMaskedLM.from_pretrained('mosaicml/mosaic-bert-candidate', use_auth_token=<your token>, trust_remote_code=True)`. For uploading to the hub, the only changes to the files that I made is to use relative imports instead of fully qualified imports. Not checking in the push to hub script for now, but lmk if you want to see it.

I tested these changes by running through the bert commands in the readme for mosaic bert.


examples/bert/src/mosaic_bert.py
```diff
@@ -88,31 +89,17 @@ def create_mosaic_bert_mlm(pretrained_model_name: str = 'bert-base-uncased',
...
-    # (otherwise, Flash Attention will be off by default)
-    if 'attention_probs_dropout_prob' not in model_config:
-        model_config['attention_probs_dropout_prob'] = 0.0
-
-    # Use `alibi_starting_size` to determine how large of an alibi tensor to
-    # create when initializing the model. You should be able to ignore
-    # this parameter in most cases.
-    if 'alibi_starting_size' not in model_config:
-        model_config['alibi_starting_size'] = 512
-
```
jacobfulano, line 101: Looks great! Thanks for cleaning this up

jacobfulano (approved):
Thanks for this straightforward implementation!

jacobfulano (approved):

PR #201, opened 2023-03-01
add no_bias option

vchiley:
[Section 2 of PaLM paper](https://arxiv.org/abs/2204.02311) shows they use "No Biases"
Screenshot 2023-03-01 at 1 52 45 PM
With this PR and adding `no_bias: true` to your config, you too can be as cool as PaLM.

For 125M model, n_params goes from `125_226_240` to `125_124_096`

Note: if you print the model, you'll see
```
            (0): FullyShardedDataParallel(
              (_fsdp_wrapped_module): FlattenParamsWrapper(
                (_fpw_module): GPTBlock(
                  (ln_1): LayerNorm((768,), eps=1e-05, elementwise_affine=True)
                  (causal_attn): FlashCausalAttention(
                    (mhsa): FlashMHA(
                      (Wqkv): Linear(in_features=768, out_features=2304, bias=False)
                      (inner_attn): FlashAttention()
                      (out_proj): Linear(in_features=768, out_features=768, bias=False)
                    )
                  )
                  (ln_2): LayerNorm((768,), eps=1e-05, elementwise_affine=True)
                  (mlp): GPTMLP(
                    (mlp_up): Linear(in_features=768, out_features=3072, bias=False)
                    (mlp_act): GELU(approximate='none')
                    (mlp_down): Linear(in_features=3072, out_features=768, bias=False)
                  )
                  (resid_attn_dropout): Dropout(p=0.0, inplace=False)
                  (resid_mlp_dropout): Dropout(p=0.0, inplace=False)
                )
              )
            )
```
where `LayerNorm((768,), eps=1e-05, elementwise_affine=True)` still has `elementwise_affine=True`. This is because LayerNorm still has its gamma multiplier.
The "model surgery" removal I do is required because you cant specify `bias=False` for LayerNorm init.


examples/llm/src/models/mosaic_gpt.py
```diff
@@ -99,6 +99,15 @@ def __init__(self, cfg: DictConfig):
         else:
             self.attn_mask = None
 
+        if cfg.get('no_bias', False):
+            for module in self.modules():
+                if isinstance(module, (nn.Linear, nn.LayerNorm)):
```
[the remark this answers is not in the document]
vchiley: image

^ this is why I did LN and Linear

examples/llm/src/models/mosaic_gpt.py
```diff
@@ -99,6 +99,15 @@ def __init__(self, cfg: DictConfig):
...
+                    if module.bias is not None:
+                        if isinstance(module.bias, nn.Parameter):
+                            print(f'Removing bias {module.bias}', end=' ')
+                            module.register_parameter('bias', None)
+                            print(f'from {module}.')
```
abhi-mosaic, line 109: +100 style points

abhi-mosaic: As much as I like this, should we gate it somehow with like `verbose: true`? Otherwise anytime we init the model there could be like 100 lines of "Removing bias..."

vchiley: Thank you

vchiley: We don't have a verbose flag being passed to model.
Should I create one or nuke the print?

examples/llm/src/models/mosaic_gpt.py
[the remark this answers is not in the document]
vchiley: Given there are no other bais in the model, I'll rm the `if isinstance(module, (nn.Linear, nn.LayerNorm)): ` check.
We can bring it back if needed, and if nothing else, it'll guard against custom layers breaking functionality (looking at you `FusedExpertNetwork` from MoE work)

examples/llm/src/models/mosaic_gpt.py
vchiley: I'll add a `verbose` flag; it'll potentially be useful in other places.

vchiley: adding `verbose` flag. It might be useful later as well.

examples/llm/src/models/mosaic_gpt.py
```diff
@@ -99,6 +99,17 @@ def __init__(self, cfg: DictConfig):
...
 
+        if cfg.get('no_bias', False):
+            for module in self.modules():
+                if hasattr(module, 'bias') and isinstance(
+                        module.bias, nn.Parameter):
+                    if cfg.get('verbose'):
+                        print(f'Removing bias ({module.bias}) from {module}.')
+                    module.register_parameter('bias', None)
+
+        if cfg.get('verbose') and cfg.get('verbose') > 2:
```
abhi-mosaic, line 110: Should it be >= 2? What secret behavior does `cfg.verbose = 2` do? 👀

abhi-mosaic (approved):
Lets try it!

examples/llm/src/models/mosaic_gpt.py
vchiley: This is me just "thinking out loud" from my undergrad days:
verbosity levels add more and more verbosity. Printing the whole model is on the extreme side of verbose, so I set it to only print if someone sets the verbose flag to >2.

vchiley: Since we're adding a verbose flag, might as well add levels  🤷‍♂️ 
or i can just rm the model print and we can default to using `verbose` as purely a bool everywhere.

PR #202, opened 2023-03-01
Rename yaml custom YAML

mvpatel2000:
The name of this YAML is a bit confusing. It's intended to be an example of customizing the repo for your model, but it doesn't explicitly call out the important change (8k seq length). We've seen multiple people (Hanlin, me) get confused by this and spend a while tracing it back, so we should rename it to be more explicit in what it does

Open to other name suggestions. This follows up a conversation offline which did not conclude on a better name


abhi-mosaic:
Could we make it `mcli-1b-max-seq-len-8k.yaml` to match the name of the config arg?

mvpatel2000:
> Could we make it mcli-1b-max-seq-len-8k.yaml to match the name of the config arg?

done

abhi-mosaic (approved):
LGTM

PR #205, opened 2023-03-02
Low precision layer norm option

dakinggg:
Verified that the model print out has the right layer for true, false and default. Verified that the first few steps of training proceed without crashing, and loss goes down.

### DIFF
--- a/examples/llm/src/models/layers/attention.py
+++ b/examples/llm/src/models/layers/attention.py
@@ -8,6 +8,8 @@ from typing import Optional
 
 import torch
 import torch.nn as nn
+from composer.algorithms.low_precision_layernorm.low_precision_layernorm import \
+    LPLayerNorm
 from einops import rearrange
 from omegaconf import DictConfig
 
@@ -120,6 +122,9 @@ class FlashCausalAttention(nn.Module):
         except ImportError as e:
             raise e
 
+        self.layernorm_class = torch.nn.LayerNorm if not cfg.get(
+            'low_precision_layernorm', False) else LPLayerNorm
+
         self.clip_qkv = cfg.get('attn_clip_qkv')
         self.attn_qk_ln = cfg.get('attn_qk_ln')
         self.d_model = cfg.d_model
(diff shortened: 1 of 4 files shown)


examples/llm/src/models/layers/attention.py
```diff
@@ -140,8 +142,10 @@ def __init__(self, cfg: DictConfig, device: Optional[str] = None):
             self.out_proj._is_residual = True  # type: ignore
 
             if self.attn_qk_ln:
-                self.q_ln = nn.LayerNorm(self.d_model, device=device)
-                self.k_ln = nn.LayerNorm(self.d_model, device=device)
+                self.layernorm_class = torch.nn.LayerNorm if not cfg.get(
```
vchiley, line 145:
```suggestion
                self.layernorm_class = nn.LayerNorm if not cfg.get(
```

we're imported nn

examples/llm/src/models/layers/attention.py
```diff
@@ -247,8 +251,10 @@ def __init__(self, cfg: DictConfig, device: Optional[str] = None):
...
```
vchiley, line 254:
```suggestion
                self.layernorm_class = nn.LayerNorm if not cfg.get(
```

we're imported nn

examples/llm/src/models/layers/gpt_blocks.py
```diff
@@ -36,9 +38,13 @@ def __init__(self,
         super().__init__()
         if cfg.get('alibi', False):
             assert cfg.attn_impl == 'triton' or cfg.attn_impl == 'torch', 'Only triton kernel or torch supports alibi'
-        self.ln_1 = nn.LayerNorm(cfg.d_model, device=device)
+
+        self.layernorm_class = torch.nn.LayerNorm if not cfg.get(
```
vchiley, line 42:
```suggestion
        self.layernorm_class = nn.LayerNorm if not cfg.get(
```

we're imported nn

examples/llm/src/models/mosaic_gpt.py
```diff
@@ -37,6 +39,9 @@ def __init__(self, cfg: DictConfig):
         else:
             raise ValueError(f'Unknown attn_impl={cfg.attn_impl}')
 
+        self.layernorm_class = torch.nn.LayerNorm if not cfg.get(
```
vchiley, line 42:
```suggestion
        self.layernorm_class = nn.LayerNorm if not cfg.get(
```

we're imported nn

vchiley (approved):
LGTM

Some nit comments
also why set `self.layernorm_class`? does it need to be attached to the obj instance?

dakinggg:
It does not need to be attached, except that it is used multiple times and I thought cleaner this way, but not a strong opinion

examples/llm/src/models/layers/attention.py
dakinggg: oops, thanks

dakinggg:
unattached it

PR #206, opened 2023-03-02
add health checker callback

dakinggg:
Verified that i get a logged statement with utils when test mode is on. Verified that throughput is ~the same for the first few batches with and without.

The health checker monitors GPU utilization and (optionally) slacks the user if one GPU has low utilization, indicating a hardware failure.


examples/common/builders.py
```diff
@@ -67,9 +71,9 @@ def build_optimizer(cfg, model):
                               weight_decay=cfg.weight_decay)
     elif cfg.name == 'decoupled_lionw':
         return DecoupledLionW(model.parameters(),
-                    lr=cfg.lr,
-                    betas=cfg.betas,
-                    weight_decay=cfg.weight_decay)
+                              lr=cfg.lr,
+                              betas=cfg.betas,
+                              weight_decay=cfg.weight_decay)
```
vchiley, line 76: sneaking in a little extra linting I see 😄

vchiley (approved):
Not exactly sure what the `HealthChecker` does but this is how you would add it.
I wouldn't be mad if you wait for other reviews tho.

examples/common/builders.py
dakinggg: i just do what the linter says :D

PR #207, opened 2023-03-02
replace speed monitor with composer one

dakinggg:
Verified that flops count is printed as the same (for 125m). 

Comparison plot for old monitor (brown) to new monitor (purple). Both runs on 4 gpus on an interactive instance, so slight difference seems expected to me.
Screen Shot 2023-03-01 at 7 18 08 PM


examples/common/builders.py
```diff
@@ -67,9 +68,9 @@ def build_optimizer(cfg, model):
                               weight_decay=cfg.weight_decay)
     elif cfg.name == 'decoupled_lionw':
         return DecoupledLionW(model.parameters(),
-                    lr=cfg.lr,
-                    betas=cfg.betas,
-                    weight_decay=cfg.weight_decay)
+                              lr=cfg.lr,
+                              betas=cfg.betas,
+                              weight_decay=cfg.weight_decay)
```
vchiley, line 73: kind of a conflict with https://github.com/mosaicml/examples/pull/206#discussion_r1122528816

vchiley (approved):
lgtm

cc @mvpatel2000 since he upstreamed SpeedMonitor to Composer

examples/llm/src/models/mosaic_gpt.py
```diff
@@ -239,7 +239,7 @@ class ComposerMosaicGPT(ComposerModel):
     def __init__(self, cfg):
         super().__init__()
         self.model = MosaicGPT(cfg)
-        self.__num_fwd_flops = None
+        self.num_fwd_flops = self.compute_num_fwd_flops()
```
mvpatel2000, line 242:
```suggestion
        self.num_fwd_flops = self._compute_num_fwd_flops()
```
Nit -- make private function?

examples/llm/src/models/mosaic_gpt.py
```diff
@@ -294,10 +294,7 @@ def add_eval_metrics(self, evaluator):
         else:
             self.eval_metrics = evaluator_metrics
 
-    @property
-    def num_fwd_flops(self):
-        if self.__num_fwd_flops:
-            return self.__num_fwd_flops
+    def compute_num_fwd_flops(self):
```
mvpatel2000, line 297:
```suggestion
    def _compute_num_fwd_flops(self):
```

examples/llm/src/models/mosaic_gpt.py
```diff
@@ -307,5 +304,11 @@ def num_fwd_flops(self):
         # there are 2 FLOPS per mac; there is A=Q*K^T and out=A*V ops (ie mult by 2)
         attn_flops_per_seq = self.model.cfg.n_layers * 2 * 2 * (
             self.model.cfg.d_model * (self.model.cfg.max_seq_len**2))
-        self.__num_fwd_flops = params_flops_per_seq + attn_flops_per_seq
-        return self.__num_fwd_flops
+        num_fwd_flops = params_flops_per_seq + attn_flops_per_seq
+
+        return num_fwd_flops
```
mvpatel2000, line 309:
```suggestion
        return params_flops_per_seq + attn_flops_per_seq
```

examples/llm/src/models/mosaic_gpt.py
```diff
@@ -307,5 +304,11 @@ def num_fwd_flops(self):
...
+
+    def flops_per_batch(self, batch):
+        # Note: this computation does not take into account padding, and assumes
+        # that the dataset has been constructed without padding
+        return self.num_fwd_flops * 3 * batch['input_ids'].shape[0]
```
mvpatel2000, line 314:
```suggestion
    def flops_per_batch(self, batch):
        # Note: this computation does not take into account padding, and assumes
        # that the dataset has been constructed without padding. Additionally, we 
        # assume the backward pass is approximately 2x the forward pass
        return self.num_fwd_flops * 3 * batch['input_ids'].shape[0]
```

mvpatel2000 (commented):
There is a print statement in `main.py` under llm examples which prints `num_fwd_flops`. Even though this new class keeps it as an attribute (so it won't crash), imo we should just remove that print statement

mvpatel2000 (approved):

dakinggg:
I'm going to leave the print because I think it is useful. Separately from this PR, we should probably switch over to actually using logging so the user can control the level (I think vitaliy just added a verbose flag which could be used here as well)

Issue #208, opened 2023-03-02
MosaicML LLM: 'key_padding_mask' is NoneType when setting "attn_impl: torch"


regnnighe:
Hi Team,

Everything worked fine with "attn_impl: flash". But when I tried to train the LLM models without the FlashAttention by setting "attn_impl: torch" in the yamls, the following error occurs.

Traceback (most recent call last):
  File "/workspace/examples_latest/examples/llm/main.py", line 215, in <module>
    main(cfg)
  File "/workspace/examples_latest/examples/llm/main.py", line 204, in main
    trainer.fit()
  File "/usr/lib/python3/dist-packages/composer/trainer/trainer.py", line 1787, in fit
    self._train_loop()
  File "/usr/lib/python3/dist-packages/composer/trainer/trainer.py", line 1950, in _train_loop
    total_loss_dict = self._train_batch(use_grad_scaling)
  File "/usr/lib/python3/dist-packages/composer/trainer/trainer.py", line 2126, in _train_batch
    optimizer.step(closure=lambda **kwargs: self._train_microbatches(
  File "/usr/lib/python3/dist-packages/torch/optim/lr_scheduler.py", line 68, in wrapper
    return wrapped(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/torch/optim/optimizer.py", line 140, in wrapper
    out = func(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/torch/autograd/grad_mode.py", line 27, in decorate_context
    return func(*args, **kwargs)
  File "/usr/lib/python3/dist-packages/composer/optim/decoupled_weight_decay.py", line 289, in step
    loss = closure()
  File "/usr/lib/python3/dist-packages/composer/trainer/trainer.py", line 2126, in <lambda>
    optimizer.step(closure=lambda **kwargs: self._train_microbatches(
  File "/usr/lib/python3/dist-packages/composer/trainer/trainer.py", line 2209, in _train_microbatches
    microbatch_loss_dict = self._train_microbatch(use_grad_scaling, current_batch_size, is_final_microbatch)
  File "/usr/lib/python3/dist-packages/composer/trainer/trainer.py", line 2255, in _train_microbatch
    self.state.outputs = self.state.model(self.state.batch)
  File "/usr/lib/python3/dist-packages/torch/nn/modules/module.py", line 1194, in _call_impl
    return forward_call(*input, **kwargs)
[...]

PR #211, opened 2023-03-03
enable torch to handle key_padding_mask=None

vchiley:
If a dataloader does not have a key_padding_mask (because all tokens are valid), the model sets `key_padding_mask=None`; torch.nn.MultiheadAttention can run with `key_padding_mask=None`.

Issue: #208


dakinggg (approved):
Can you add a simple test that would've caught this?

vchiley:
tests added
verified that test fails without updt and passes with updt

PR #217, opened 2023-03-09
Removes `add_eval_metrics` and just puts them in the relevant constructors

dakinggg:
Manually tested by running mosaic gpt and gpt neo and confirming that it made it through the first eval with icl metrics.

### DIFF
--- a/examples/llm/src/models/hf/hf_t5.py
+++ b/examples/llm/src/models/hf/hf_t5.py
@@ -5,10 +5,7 @@
 
 from __future__ import annotations
 
-from composer.metrics.nlp import (InContextLearningLMAccuracy,
-                                  InContextLearningMetric,
-                                  InContextLearningMultipleChoiceAccuracy,
-                                  LanguageCrossEntropy, MaskedAccuracy)
+from composer.metrics.nlp import LanguageCrossEntropy, MaskedAccuracy
 from omegaconf import DictConfig
 from transformers import AutoConfig, AutoTokenizer, T5ForConditionalGeneration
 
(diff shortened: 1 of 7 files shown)


examples/llm/src/models/hf/hf_t5.py
```diff
@@ -81,13 +84,21 @@ def __init__(self, cfg: DictConfig):
                                  ignore_index=_HF_IGNORE_INDEX),
             MaskedAccuracy(ignore_index=_HF_IGNORE_INDEX)
         ]
+        eval_metrics = [
+            LanguageCrossEntropy(vocab_size=vocab_size,
+                                 ignore_index=_HF_IGNORE_INDEX),
+            MaskedAccuracy(ignore_index=_HF_IGNORE_INDEX),
+            InContextLearningLMAccuracy(),
+            InContextLearningMultipleChoiceAccuracy()
```
alextrott16, line 92: These ICL eval tools don't support encoder-decoder yet. Maybe we should hold off on adding these until they are supported?

alextrott16 (approved):
LGTM and if it works then that seems like a reasonable test.

Out of curiosity, how does this work with eval dataloaders that don't specify the metrics that apply to them? I know that _evaluators_ let you name the relevant metrics, but that's not required, right? I thought that if you don't use an evaluator object for the eval dataloader, it will try to compute all the eval metrics. And I'd expect that to be an issue when, say, your standard LM eval dataloader doesn't produce batches that work with one of the ICL metrics.

Can you help correct my misunderstanding?

dakinggg:
Your understanding is correct. In examples repo we make sure all eval things are `Evaluator`, so just hardcoding the metrics seemed like the cleanest solution here. This is actually just necessary given the current composer design for eval metrics, which is a bit wonky. We do have some code in `update_metric` that skips updating metrics when the batch is the wrong thing, so it still shouldn't crash in that case, but you do get some bad logging I believe, don't quite remember.

examples/llm/src/models/hf/hf_t5.py
dakinggg: ah, good call. will remove.

PR #218, opened 2023-03-09
track loss finite diff

vchiley:
callback to plot `metric_{i} - metric_{i-1}`.
This is effectively a finite diff approx of the derivative of the loss which might help us see which networks / configurations have plateaued / converged and which are still "training" relatively fast.

Example [here](https://wandb.ai/mosaic-ml/test_fdiff_loss)

Given train loss is high variance, I'm not sure this should be used for train metrics, but it could be useful for eval metrics.

Eval metrics and fdiff:
Screenshot 2023-03-09 at 9 43 59 AM

A positive fdiff means loss is diverging; negative means converging; 0 means model has converged.
In isolation this means nothing, but is useful for comparisons.


examples/common/fdiff_loss.py
```diff
@@ -0,0 +1,70 @@
...
+        return {
+            'train_prev_loss': self.train_prev_loss,
+            'train_prev_metric': self.train_prev_metric,
+            'eval_prev_metric': self.eval_prev_metric,
+        }
+
+    def load_state_dict(self, state: Dict[str, Any]) -> None:
+        self.train_prev_loss = state['train_prev_loss']
+        self.train_prev_metric = state['train_prev_metric']
+        self.eval_prev_metric = state['eval_prev_metric']
```
abhi-mosaic, line 38: I would strongly consider how much value it adds to have state_dict and resumption (saves you from waiting 1 train batch or 1 eval period).

Because if anything about this callback's state_dict ever changes, it would break checkpoint compatibility. So we would probably be committing to make all future versions of this callback backwards-compatible.

(fyi this is why we ripped almost everything out of `SpeedMonitor.state_dict`. Stateless = no bugs :)

examples/common/builders.py
```diff
@@ -27,6 +28,8 @@ def build_callback(name, kwargs):
         return SpeedMonitorMFU(window_size=kwargs.get('window_size', 1),
                                gpu_flops_available=kwargs.get(
                                    'gpu_flops_available', None))
+    elif name == 'fdiff_loss':
+        return FDiffLoss()
```
abhi-mosaic, line 32: Since this is diff-ing all metrics, why not just `fdiff` and `FDiffMetrics` or something like that?

abhi-mosaic (approved):
LGTM, but see my comment about state_dict

examples/common/fdiff_loss.py
vchiley: I guess if we restart, we'll only be missing one itr idx in the fdiff graph; it wont even be noticeable.
Ok going to make it stateless.

PR #220, opened 2023-03-09
enable flexible emb init

vchiley:
allows user to init embed layer more flexibly.

eg: setting `emb_init_std: 1` will override init_fn_ to init embed layer using normal distribution with mean=0 and std=1


dakinggg (approved):
LGTM, would be good to add a basic test, like if you set the things to 0, you get the expected warning and the embedding layers is 0s. to at least exercise the new code paths

PR #221, opened 2023-03-09
enabling logit scaling

vchiley:
enables scaling output logits; similar to a softmax "temperature"
the [PaLM paper](https://arxiv.org/abs/2204.02311) uses scale 1/sqrt(cfg.d_model) (see Sec 5 Weight initialization)

Implantation notes: 
[F.cross_entropy](https://pytorch.org/docs/stable/generated/torch.nn.functional.cross_entropy.html#torch-nn-functional-cross-entropy) does not enable temperature.
Since we'd have to add it to the model anyway, I just made it a buffer so that it'd be part of the model state_dict.

### DIFF
--- a/examples/llm/src/models/mosaic_gpt.py
+++ b/examples/llm/src/models/mosaic_gpt.py
@@ -192,7 +192,12 @@ class MosaicGPT(nn.Module):
         # enables scaling output logits; similar to a softmax "temperature"
         # PaLM paper uses scale 1/sqrt(cfg.d_model)
         if self.cfg.get('logit_scale') is not None:
-            logits *= self.cfg.get('logit_scale')
+            logit_scale = self.cfg.get('logit_scale')
+            if logit_scale == 0:
+                warnings.warn(
+                    f'Multiplying logic scale by {logit_scale=}. This will produce degenerate probabilities.'
+                )
+            logits *= logit_scale
 
         return logits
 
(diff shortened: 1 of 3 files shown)


examples/llm/src/models/mosaic_gpt.py
```diff
@@ -92,6 +92,13 @@ def __init__(self, cfg: DictConfig):
         self.transformer.update(
             {'ln_f': nn.LayerNorm(cfg.d_model, device=cfg.init_device)})
 
+        # enables scaling output logits; similar to a softmax "temperature"
+        # PaLM paper uses scale 1/sqrt(cfg.d_model)
+        if cfg.get('logit_scale'):
+            self.register_buffer('logit_scale', torch.Tensor([cfg.logit_scale]))
+        else:
+            self.register_buffer('logit_scale', None)
+
```
abhi-mosaic, line 101: From your comment I guess this is in order to save `logit_scale` in the model checkpoint, but it could also just live as part of the `model_cfg` and then we would just use a float rather than a Tensor?

Thinking about this in relation to `alibi_bias` or `softmax_scale` or other args like that, should we just keep it as a `cfg` float for now? I agree we need a better solution for saving model configs colocated with the checkpoint weights, but not sure if we should do a mix of solutions.

vchiley: ok I can leave it as just part of `self.cfg`. 
I was just trying to be thorough.

samhavens (approved):
LGTM

PR #224, opened 2023-03-09
Small and Wang initializations

samhavens:
This adds type signatures to the functions and fixes some spelling mistakes. The main contribution is to add 2 initialization schemes: Small Init, from Transformers without Tears, and Wang Init, which GPT NeoX uses in combination with Small Init.

I would like to test these out but need to coordinate with the Hero Runners.


samhavens:
@vchiley's feedback (copied from Slack):
> i'd modify it to just add generic normal_init_ and uniform_init_ init fns since we don't have those yet
> then you can effectively play around with inits like "small" with out needing to do anything else.
> but yeah like i said, small seems very similar to kaiming_normal_ (with a diff "gain") for all but one of the layers in the gpt block (1 out of 6) and I think we can use init_div_is_residual to fix that one

PR #226, opened 2023-03-10
Parametrized init div is residual, NeoX initialization

samhavens:
This PR:

* fixes a few spelling mistakes
* allows the user to specify a particular value for `init_div_is_residual`, though if not specified it still defaults to the value it has always had, and it still respects `init_div_is_residual: false`
* adds `_normal_init_` and `_normal_param_init_fn_` fns and defines `baseline_param_init_fn_` in terms of them
* adds `small_param_init_fn_` from Transformers without Tears 
image
* uses `small_param_init_fn_` and the now-configurable `init_div_is_residual` to implement the initialization scheme for Neox 20B

### DIFF
--- a/examples/llm/src/models/param_init_fns.py
+++ b/examples/llm/src/models/param_init_fns.py
@@ -50,14 +50,15 @@ def generic_param_init_fn_(module, cfg, init_fn_):
     init_div_is_residual = cfg.get('init_div_is_residual', True)
 
     if init_div_is_residual is False:
-        div_is_residual = 1
-    if init_div_is_residual is True:
+        pass
+    elif init_div_is_residual is True:
         div_is_residual = math.sqrt(2 * cfg.n_layers)
     elif isinstance(init_div_is_residual, float) or isinstance(
             init_div_is_residual, int):
         div_is_residual = init_div_is_residual
     elif isinstance(init_div_is_residual,
                     str) and init_div_is_residual.isnumeric():
+        # do not trust YAML parsing to always convert numbers to numbers
         div_is_residual = float(init_div_is_residual)
     else:
         raise ValueError(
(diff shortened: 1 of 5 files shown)


examples/llm/src/models/param_init_fns.py
```diff
@@ -153,16 +176,46 @@ def generic_param_init_fn_(module, cfg, init_fn_):
...
 
     if cfg.get('verbose', 0) > 1:
         warnings.warn(
-            f'Using torch.nn.init.normal_ init fn mean=0.0, std={cfg.init_std}')
+            f'Using torch.nn.init.normal_ init fn mean=0.0, std={std}')
 
     generic_param_init_fn_(module, cfg, init_fn_)
 
 
+def baseline_param_init_fn_(module, cfg):
```
vchiley, line 193: since we out here, might as well allow `normal_` where we allow user to pass in both mean and std?
also enable `uniform_` if you're feeling up to it.

samhavens: so, the initialization fn will always be called `MODEL_INIT_REGISTRY[init_fn_name](module, self.cfg)`... so the only way to pass mean and std are through the cfg (i.e. not referencing other values in the cfg; `n_layers`)... so this request is equivalent to looking for `cfg.init_mean` (plus adding uniform, which I think should wait till someone wants it)?

vchiley: yeah mean would be passed as part of `cfg.init_mean`
we can leave it out till someone wants it.

examples/llm/src/models/param_init_fns.py
```diff
@@ -45,12 +45,33 @@ def generic_param_init_fn_(module, cfg, init_fn_):
...
-    # enable user to divide _is_residual weights by math.sqrt(2 * cfg.n_layers)
-    div_is_residual = cfg.get('init_div_is_residual', True)
-    if div_is_residual:
+    # enable user to divide _is_residual weights by
+    # a value which defaults to math.sqrt(2 * cfg.n_layers)
+    init_div_is_residual = cfg.get('init_div_is_residual', True)
+
+    if init_div_is_residual is False:
+        # not used, for pyright
+        div_is_residual = 1.0
```
vchiley, line 54: :old-man-yells-at-pytest:

examples/llm/src/models/param_init_fns.py
```diff
@@ -153,16 +176,46 @@ def generic_param_init_fn_(module, cfg, init_fn_):
...
     if cfg.get('verbose', 0) > 1:
         warnings.warn(
-            f'Using torch.nn.init.normal_ init fn mean=0.0, std={cfg.init_std}')
+            f'Using torch.nn.init.normal_ init fn mean=0.0, std={std}')
 
     generic_param_init_fn_(module, cfg, init_fn_)
 
 
+def baseline_param_init_fn_(module, cfg):
+    _normal_param_init_fn_(module, cfg, std=cfg.init_std)
```
vchiley, line 194: do `get` then cleanly error out with informative Error instead?

vchiley (approved):
lgtm

(with small nit)

examples/llm/src/models/param_init_fns.py
samhavens: good call, updated

PR #229, opened 2023-03-10
Prepare model for inference serving

dskhudia:
We wrap the model with a generator and provide APIs to get the tokenizer and generator. The model is loaded from a checkpoint. 

Some future improvements:

Inference optimizations such as KV caching
Deepspeed-based optimization

### DIFF
--- a/examples/llm/inference/inference.py
+++ b/examples/llm/inference/inference.py
@@ -1,0 +1,122 @@
+# Copyright 2022 MosaicML Examples authors
+# SPDX-License-Identifier: Apache-2.0
+import warnings
+from typing import List
+
+import torch
+from composer.core import get_precision_context
+from omegaconf import OmegaConf as om
+
+from examples.llm.src import COMPOSER_MODEL_REGISTRY, TOKENIZER_REGISTRY
+
+
+def sample_top_p(probs, p):
+    probs_sort, probs_idx = torch.sort(probs, dim=-1, descending=True)
+    probs_sum = torch.cumsum(probs_sort, dim=-1)
+    mask = probs_sum - probs_sort > p
+    probs_sort[mask] = 0.0
+    probs_sort.div_(probs_sort.sum(dim=-1, keepdim=True))
+    next_token = torch.multinomial(probs_sort, num_samples=1)
+    next_token = torch.gather(probs_idx, -1, next_token)
+    return next_token
+
+
+class MosaicGPTInference:
+
+    def __init__(self, cfg, model, tokenizer):
+        self.cfg = cfg
+        self.model = model
+        self.tokenizer = tokenizer
+
+    def generate(
+        self,
+        prompts: List[str],
+        max_gen_len: int,
+        temperature: float = 0.8,
+        top_p: float = 0.95,
+    ) -> List[str]:
+        bsz = len(prompts)
+
+        prompt_tokens = [self.tokenizer.encode(x) for x in prompts]
+
+        min_prompt_size = min([len(t) for t in prompt_tokens])
+        max_prompt_size = max([len(t) for t in prompt_tokens])
+
+        total_len = min(self.cfg.max_seq_len, max_gen_len + max_prompt_size)
+
+        tokens = torch.full((bsz, total_len), -100).cuda().long()
+        for k, t in enumerate(prompt_tokens):
+            tokens[k, :len(t)] = torch.tensor(t).long()
+        input_text_mask = tokens != -100
+        start_pos = min_prompt_size
+        for cur_pos in range(start_pos, total_len):
+            with torch.no_grad():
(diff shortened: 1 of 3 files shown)


dskhudia:
@dakinggg any idea why lint is failing on this?

dakinggg:
pyright error
```
  /home/runner/work/examples/examples/examples/llm/inference/inference.py:109:25 - error: Argument of type "Literal['load_path']" cannot be assigned to parameter "index" of type "int" in function "get"
    "Literal['load_path']" is incompatible with "int" (reportGeneralTypeIssues)
```

dskhudia:
Thanks @dakinggg , Not sure how I missed the error but I did search for the errors in logs. :-(

examples/llm/inference/inference.py
```diff
@@ -0,0 +1,122 @@
...
+    return tokenizer
+
+
+def get_mosaicgpt_inference_model(checkpoint_yaml_path: str, tokenizer):
+    with open(checkpoint_yaml_path) as f:
+        cfg = om.load(f)
+    # set init_device to cpu for checkpoint loading
+    # ToDo: Directly load a checkpoint into 'meta' model
+    cfg.model.init_device = 'cpu'
+    model = build_composer_model(cfg.model)
```
abhi-mosaic, line 107: @dakinggg @dskhudia At a high level, do we intend to serve the `ComposerModel` wrapper at inference time, or just the PyTorch module `MosaicGPT`? Thinking ahead, if `MosaicGPT` is a HF model, we would serve that rather than the `ComposerHuggingFace` wrapper, right?

dskhudia: Once MosaicGPT is a HF model, we would use the GenereteMixIn. @dakinggg is working on this integration. This PR is a stopgap solution for serving our trained checkpoints using our inference solution.

dakinggg: I think it depends a little bit on what we do for parallelism...but yes, probably we would just serve the HF model.

dskhudia: Either way (MosaicGPT or HF), for parallelism, we can currently pass the model to deepspeed.

examples/llm/inference/inference.py
```diff
@@ -0,0 +1,122 @@
...
+            if temperature > 0:
+                probs = torch.softmax(logits / temperature, dim=-1)
+                next_token = sample_top_p(probs, top_p)
+            else:
+                next_token = torch.argmax(logits, dim=-1)
+            next_token = next_token.reshape(-1)
+            # only replace token if prompt has already been generated
+            next_token = torch.where(input_text_mask[:, cur_pos],
+                                     tokens[:, cur_pos], next_token)
+            tokens[:, cur_pos] = next_token
```
samhavens, line 69: do we want to break out of the `for` loop if `next_token == self.tokenizer.eos_token_id` here?

samhavens (commented):
IIUC this is a 0->1 PR (adding inference capabilities) so my default is to be permissive in approving. Also I believe this is a stopgap, so I don't want to slow it down.

I think we can improve generation speed by stopping as soon as we generate the EOS token.

Also, there are tasks where we want `generate` functionality during training; so do we have thoughts on if we want this class to be a mixin and to name it something like `SimpleGenerationMixin` and to locate it in the main `llm` code? I am okay with this change coming later if we decide it is warranted, just wanted to surface it

examples/llm/inference/inference.py
samhavens: Oops, I was only thinking about batch size = 1 here

samhavens: and my other point, about deeper integration, will be resolved with Daniel's work. So I am resolving this

samhavens (approved):
I don't see any blocking issues. I believe this code will be removed when @dakinggg adds more thorough generation soon anyway, so I am approving.

examples/llm/inference/inference.py
```diff
@@ -0,0 +1,122 @@
...
+    # set init_device to cpu for checkpoint loading
+    # ToDo: Directly load a checkpoint into 'meta' model
+    cfg.model.init_device = 'cpu'
+    model = build_composer_model(cfg.model)
+
+    ckpt_load_path = cfg.get('load_path', None)  # type: ignore
+    if ckpt_load_path is None:
+        raise ValueError('Checkpoint load_path is required for exporting.')
+
+    checkpoint = torch.load(ckpt_load_path, map_location='cpu')
```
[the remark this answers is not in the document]
dskhudia: There is no check for the validity of the checkpoint. This checkpoint is used provided so the usual caveat of only load the checkpoints from the sources you trust applies. Also this inference path is without export (.i.e., we are not using onnx/torchscript).

PR #230, opened 2023-03-10
updating triton attn mask doc str

vchiley:
docstring updt
https://mosaicml.atlassian.net/browse/RESEARCH-544

### DIFF
--- a/examples/llm/src/models/layers/flash_attention.py
+++ b/examples/llm/src/models/layers/flash_attention.py
@@ -19,6 +19,7 @@ class FlashAttention(nn.Module):
     """Implement the scaled dot product attention with softmax.
 
     Args:
+        num_heads: number of attention heads
         softmax_scale: The temperature to use for the softmax attention.
                       (default: 1/sqrt(d_keys) where d_keys is computed at
                       runtime)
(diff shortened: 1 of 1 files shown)


examples/llm/src/models/layers/flash_attention.py
```diff
@@ -58,14 +58,7 @@ def forward(
...
-            attn_mask: If specified, a 2D or 3D mask preventing attention to certain positions. Must be of shape
-                :math:`(L, S)` or :math:`(N\cdot\text{num\_heads}, L, S)`, where :math:`N` is the batch size,
-                :math:`L` is the target sequence length, and :math:`S` is the source sequence length. A 2D mask will be
-                broadcasted across the batch while a 3D mask allows for a different mask for each entry in the batch.
-                Binary, byte, and float masks are supported. For a binary mask, a ``True`` value indicates that the
-                corresponding position is not allowed to attend. For a byte mask, a non-zero value indicates that the
-                corresponding position is not allowed to attend. For a float mask, the mask values will be added to
-                the attention weight.
+            attn_mask: If specified, a 4D mask of floats which will be added to the attention weight. Must braodcast to (B, H, S, S).
             is_causal: If specified, applies a causal mask as attention mask. Mutually exclusive with providing attn_mask.
```
alextrott16, line 62: Can you confirm that `is_causal` is still mutually exclusive with providing `attn_mask`? Some of our code makes me think it's not.

alextrott16 (changes requested):

examples/llm/src/models/layers/flash_attention.py
```diff
@@ -58,14 +58,7 @@ def forward(
                 Binary and byte masks are supported.
```
alextrott16, line 58: (Sorry this comment was supposed to be part of my review but it didn't take)

Is this a good time to axe `key_padding_mask` from this module? At the very least we should also update the text here because it implies that we're using this argument, but we're not because we're expecting `key_padding_mask` was already baked into `attn_mask`.

If you're good with axing stuff, `need_weights` and `average_attn_weights` are looking mighty deprecated too.

examples/llm/src/models/layers/flash_attention.py
vchiley: which code?

vchiley: In the [triton code](https://github.com/HazyResearch/flash-attention/blob/main/flash_attn/flash_attn_triton.py#L143), you see that first causal is done, then bias is added

vchiley: I think that mutually exclusive comment is ripped from somewhere random...

alextrott16: Yeah, because our alibi implementation was letting us both set `attn_mask` and have `is_causal=True`, right? So they can be mixed and matched?

examples/llm/src/models/layers/flash_attention.py
vchiley: I'll rm `need_weights` and `average_attn_weights` but keep `key_padding_mask` since it a very common arg. Keeping it enables me to throw an informative error.

examples/llm/src/models/layers/flash_attention.py
vchiley: thank you for catching that.

alextrott16 (approved):
LGTM! Thanks for addressing this!

PR #210, opened 2023-03-02
Add Optimized GPT

nik-mosaic:
Adds code for an optimized version of MosaicGPT. The user can specify cfg.gpt_block = 'optimized' to run this code, and cfg.gpt_block = 'standard' (or omit it) to run the standard code.

This PR contains:
  - In `src/mosaic_gpt.py` and `src/models/layers/gpt_blocks.py`: Modeling changes to enable the optimized GPT: moving first layernorm of GPT blocks 2 - n to the end of blocks 1 - n-1, and replacing the final layernorm with `ln_f` with `ln_i`. These should be math-equivalent, so if you choose not to use cfg.gpt_block = 'optimized', your model will be unchanged.
  - In `src/models/layers/gpt_blocks.py`: The new OptimizedGPTBlock itself.
  - In `src/mosaic_gpt.py`: An option to use Fused Cross Entropy, which is enabled by default, installed via the standard `pip install .[llm]`. Otherwise, you can set `cfg.loss_fn=torch_crossentropy` to disable this and use standard torch.nn.CrossEntropy().
  - In the `csrc/` folder: C++/CUDA code for the one custom fusion we include. Since the HazyResearch DropoutAddLayerNorm does not support 30B and 70B models, we add code to support these models and require the user install it.
- In `llm/README.md` and `llm/csrc/README.md`: Installation instructions for dependencies for optimized MosaicGPT.

Along for the ride:
- Changes all vocab_sizes in YAMLs to 50304 (`icl_evals/yamls`, `yamls/mosaic_gpt/`, `mcloud/`)

Future Work:
- FusedMLP has been removed since it doesn't help performance on bf16 for CUDA < 11.8. In the future, when we support PyTorch 2.0 + CUDA 11.8, I will retest this.
- After this PR, I will make a pull request in the HazyResearch repo to add DropoutAddLayerNorm support for 30B and 70B model sizes . Then, we can remove the `csrc/` folder and include DropoutAddLayerNorm as a single line in a requirements file.
- Bake in DropoutAddLayerNorm into our docker images, like we've done with FlashAttention, so that it does not need to build every time (~20 min. build time).

PR #231, opened 2023-03-11
Implement DeepNorm

samhavens:
## DeepNorm!

This PR:
* adds `GPTBlockPostLN`, which is a PostLN block (we have been using PreLN) 
image
* adds a config, `model.norm_location`, which defaults to `pre_ln`, but if is set otherwise, causes us to use `GPTBlockPostLN` instead of `GPTBlock`
* adds a new initialization registry value for the config `param_init_fn`, `deepnorm_`, which uses DeepNorm initialization from the [DeepNet](https://arxiv.org/abs/2203.00555) paper
* if `param_init_fn: deepnorm_` then we also use `GPTBlockPostLN`

### DIFF
--- a/examples/llm/src/models/layers/gpt_blocks.py
+++ b/examples/llm/src/models/layers/gpt_blocks.py
@@ -23,7 +23,7 @@ class GPTMLP(nn.Module):
                                   device=device)
         self.mlp_down._is_residual = True  # type: ignore
 
-    def forward(self, x):
+    def forward(self, x: torch.Tensor):
         return self.mlp_down(self.mlp_act(self.mlp_up(x)))
 
 
@@ -31,7 +31,7 @@ class GPTBlock(nn.Module):
 
     def __init__(self,
                  cfg: DictConfig,
-                 causal_attn_cls,
+                 causal_attn_cls: nn.Module,
                  device: Optional[str] = None):
         super().__init__()
         if cfg.get('alibi', False):
@@ -55,4 +55,40 @@ class GPTBlock(nn.Module):
         m = self.ln_2(x)
         n = self.mlp(m)
         x = x + self.resid_mlp_dropout(n)
+        return x
+
+
+class GPTBlockPostLN(nn.Module):
+
+    def __init__(self,
+                 cfg: DictConfig,
+                 causal_attn_cls: nn.Module,
+                 device: Optional[str] = None):
+        super().__init__()
+        if cfg.get('alibi', False):
+            assert cfg.attn_impl == 'triton' or cfg.attn_impl == 'torch', 'Only triton kernel or torch supports alibi'
+        self.ln_1 = nn.LayerNorm(cfg.d_model, device=device)
+        self.causal_attn = causal_attn_cls(cfg, device)
+        self.ln_2 = nn.LayerNorm(cfg.d_model, device=device)
+        self.mlp = GPTMLP(cfg, device=device)
+        self.resid_attn_dropout = nn.Dropout(cfg.resid_pdrop)
+        self.resid_mlp_dropout = nn.Dropout(cfg.resid_pdrop)
+        if cfg.get('norm_style', 'post_ln') == 'deepnorm':
+            # constant for decoder-only models
+            # see the paper DeepNet, https://arxiv.org/abs/2203.00555
+            self.residual_scale = (2 * cfg.n_layers)**0.25
+        else:
+            self.residual_scale = 1.0
+
+    def forward(
+        self,
+        x: torch.Tensor,
+        key_padding_mask: Optional[torch.ByteTensor] = None,
+        attn_mask: Optional[torch.Tensor] = None,
+    ) -> torch.Tensor:
+        a, _ = self.causal_attn(x, key_padding_mask, attn_mask)
+        residue = x * self.residual_scale
+        x = self.ln_1(residue + self.resid_attn_dropout(a))
+        residue = x * self.residual_scale
+        x = self.ln_2(residue + self.resid_mlp_dropout(self.mlp(x)))
         return x
--- a/examples/llm/src/models/layers/gpt_blocks.py
+++ b/examples/llm/src/models/layers/gpt_blocks.py
@@ -73,7 +73,7 @@ class GPTBlockPostLN(nn.Module):
         self.mlp = GPTMLP(cfg, device=device)
         self.resid_attn_dropout = nn.Dropout(cfg.resid_pdrop)
         self.resid_mlp_dropout = nn.Dropout(cfg.resid_pdrop)
-        if cfg.get('norm_style', 'post_ln') == 'deepnorm':
+        if cfg.get('param_init_fn', 'default_') == 'deepnorm_':
             # constant for decoder-only models
             # see the paper DeepNet, https://arxiv.org/abs/2203.00555
             self.residual_scale = (2 * cfg.n_layers)**0.25
--- a/examples/llm/src/models/layers/gpt_blocks.py
+++ b/examples/llm/src/models/layers/gpt_blocks.py
@@ -31,7 +31,7 @@ class GPTBlock(nn.Module):
 
     def __init__(self,
                  cfg: DictConfig,
-                 causal_attn_cls: nn.Module,
+                 causal_attn_cls,
                  device: Optional[str] = None):
         super().__init__()
         if cfg.get('alibi', False):
@@ -62,7 +62,7 @@ class GPTBlockPostLN(nn.Module):
 
     def __init__(self,
                  cfg: DictConfig,
-                 causal_attn_cls: nn.Module,
+                 causal_attn_cls,
                  device: Optional[str] = None):
         super().__init__()
         if cfg.get('alibi', False):
--- a/examples/llm/src/models/mosaic_gpt.py
+++ b/examples/llm/src/models/mosaic_gpt.py
@@ -69,7 +69,11 @@ class MosaicGPT(nn.Module):
         # default to PreLN blocks for consistency for now
         # if PostLN ends up being better, switch this default
         gpt_block_type = gpt_blocks.GPTBlock if cfg.get(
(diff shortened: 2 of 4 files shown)


examples/llm/src/models/layers/gpt_blocks.py
```diff
@@ -56,3 +56,39 @@ def forward(
...
+                 cfg: DictConfig,
+                 causal_attn_cls: nn.Module,
+                 device: Optional[str] = None):
+        super().__init__()
+        if cfg.get('alibi', False):
+            assert cfg.attn_impl == 'triton' or cfg.attn_impl == 'torch', 'Only triton kernel or torch supports alibi'
+        self.ln_1 = nn.LayerNorm(cfg.d_model, device=device)
+        self.causal_attn = causal_attn_cls(cfg, device)
+        self.ln_2 = nn.LayerNorm(cfg.d_model, device=device)
+        self.mlp = GPTMLP(cfg, device=device)
```
vchiley, line 73:
```suggestion
        self.causal_attn = causal_attn_cls(cfg, device)
        self.ln_1 = nn.LayerNorm(cfg.d_model, device=device)
        self.mlp = GPTMLP(cfg, device=device)
        self.ln_2 = nn.LayerNorm(cfg.d_model, device=device)

```

nit: I like to init things in the order they are applied.

examples/llm/src/models/param_init_fns.py
```diff
@@ -293,9 +293,116 @@ def xavier_normal_param_init_fn_(module, cfg):
...
+                    warnings.warn(f'Embedding layer initialized to 0.')
+                lim = [-lim, lim]
+            a, b = lim
+            emb_init_fn_ = partial(torch.nn.init.uniform_, a=a, b=b)
+            if cfg.get('verbose', 0) > 1:
+                warnings.warn(
+                    f'Embedding layer initialized using uniform distribution in range {lim}.'
+                )
+        else:
+            emb_init_fn_ = fvo_init_fn_  # deepnet paper does not mention embedding init
```
vchiley, line 346: if it does not mention it, there is a good change they are using normal_(mean=0, std=1) since that is the pytorch default init for emb

samhavens: thanks, updated!

examples/llm/src/models/mosaic_gpt.py
```diff
@@ -66,6 +66,17 @@ def __init__(self, cfg: DictConfig):
...
+        # default to PreLN blocks for consistency for now
+        # if PostLN ends up being better, switch this default
+        # user can specify using post_ln
+        gpt_block_type = gpt_blocks.GPTBlock if cfg.get(
+            'norm_location',
+            'pre_ln') == 'pre_ln' else gpt_blocks.GPTBlockPostLN
+        # or, if the initialization is DeepNorm, know we have to use PostLN
+        gpt_block_type = gpt_blocks.GPTBlockPostLN if cfg.get(
+            'param_init_fn',
+            'baseline_') == 'deepnorm_' else gpt_blocks.GPTBlock
```
vchiley, line 78:
```suggestion
        # if the initialization is DeepNorm, know we have to use PostLN
        if cfg.get('param_init_fn', 'baseline_') == 'deepnorm_':
            raise NotImplementedError(
                f'Deepnorm init is only implemented for PostLN block. Set `norm_location: post_ln ` in model configuration.'
            )
```

Silent model change is danger...

examples/llm/src/models/layers/gpt_blocks.py
```diff
@@ -56,3 +56,40 @@ def forward(
...
+        self.ln_2 = nn.LayerNorm(cfg.d_model, device=device)
+
+        self.resid_attn_dropout = nn.Dropout(cfg.resid_pdrop)
+        self.resid_mlp_dropout = nn.Dropout(cfg.resid_pdrop)
+        if cfg.get('param_init_fn', 'default_') == 'deepnorm_':
+            # constant for decoder-only models
+            # see the paper DeepNet, https://arxiv.org/abs/2203.00555
+            self.residual_scale = (2 * cfg.n_layers)**0.25
+        else:
+            self.residual_scale = 1.0
```
vchiley, line 82:
```suggestion
        self.residual_scale = None
        if cfg.get('param_init_fn', 'default_') == 'deepnorm_':
            # constant for decoder-only models
            # see the paper DeepNet, https://arxiv.org/abs/2203.00555
            self.residual_scale = (2 * cfg.n_layers)**0.25
```

examples/llm/src/models/layers/gpt_blocks.py
```diff
@@ -56,3 +56,40 @@ def forward(
...
+
+    def forward(
+        self,
+        x: torch.Tensor,
+        key_padding_mask: Optional[torch.ByteTensor] = None,
+        attn_mask: Optional[torch.Tensor] = None,
+    ) -> torch.Tensor:
+        a, _ = self.causal_attn(x, key_padding_mask, attn_mask)
+        residue = x * self.residual_scale
```
vchiley, line 91:
```suggestion
        residue = x * self.residual_scale if self.residual_scale else x
```

skips a memory bound mult op if not needed

examples/llm/src/models/layers/gpt_blocks.py
```diff
@@ -56,3 +56,40 @@ def forward(
...
+        x = self.ln_1(residue + self.resid_attn_dropout(a))
+        residue = x * self.residual_scale
```
vchiley, line 93:
```suggestion
        residue = x * self.residual_scale if self.residual_scale else x
```

skips a memory bound mult op if not needed

examples/llm/src/models/param_init_fns.py
```diff
@@ -293,9 +293,99 @@ def xavier_normal_param_init_fn_(module, cfg):
...
+        elif cfg.get('emb_init_uniform_lim') is not None:
+            warnings.warn(
+                f'emb_init_uniform_lim ignored when using param_init_fn: deepnorm_'
+            )
+
+        emb_init_fn_(module.weight)
+        if cfg.get('verbose', 0) > 1:
+            warnings.warn(
+                'Embedding layer initialized using normal distribution with mean=0 and std=1.'
+            )
```
vchiley, line 331:
```suggestion
        if cfg.get('emb_init_std') is not None:
            std = cfg.get('emb_init_std')
            if std == 0:
                warnings.warn(f'Embedding layer initialized to 0.')
            emb_init_fn_ = partial(torch.nn.init.normal_, mean=0.0, std=std)
            if cfg.get('verbose', 0) > 1:
                warnings.warn(
                    f'Embedding layer initialized using normal distribution with mean=0 and {std=}.'
                )
        elif cfg.get('emb_init_uniform_lim') is not None:
            lim = cfg.get('emb_init_uniform_lim')
            if isinstance(lim, Sequence):
                if len(lim) > 2:
                    raise ValueError(
                        f'Uniform init requires a min and a max limit. User input: {lim}.'
                    )
                if lim[0] == lim[1]:
                    warnings.warn(f'Embedding layer initialized to {lim[0]}.')
            else:
                if lim == 0:
                    warnings.warn(f'Embedding layer initialized to 0.')
                lim = [-lim, lim]
            a, b = lim
            emb_init_fn_ = partial(torch.nn.init.uniform_, a=a, b=b)
            if cfg.get('verbose', 0) > 1:
                warnings.warn(
                    f'Embedding layer initialized using uniform distribution in range {lim}.'
                )
        else:
            emb_init_fn_ = partial(torch.nn.init.normal_, mean=0.0, std=1)
            if cfg.get('verbose', 0) > 1:
                warnings.warn(
                    'Embedding layer initialized using normal distribution with mean=0 and std=1.'
                )
        emb_init_fn_(module.weight)
```

Given '`DeepNet paper doesn't mention what they used, so assume PyTorch default `' is an assumption, we should default to it, but allow the user to specify what they want incase we find out that our assumption is off.

examples/llm/src/models/param_init_fns.py
```diff
@@ -293,9 +293,99 @@ def xavier_normal_param_init_fn_(module, cfg):
...
+    beta = (8 * cfg.n_layers)**-0.25
+    fvo_init_fn_ = partial(nn.init.xavier_normal_, gain=beta)
+    qk_init_fn_ = partial(nn.init.xavier_normal_, gain=1.)
+
+    if isinstance(module, nn.Linear):
+        # Linear
+        if hasattr(module, '_fused'):
+            fused_init_helper_(module, fvo_init_fn_)
+        else:
+            fvo_init_fn_(module.weight)
```
vchiley, line 311: Note:
Later on when we init
```
    elif isinstance(module, nn.MultiheadAttention):
        # torch's MultiheadAttention
```
this is only initializing `torch.nn.MultiheadAttention`.
other layers implementations just use linear for Wqkv (which is a fused layer) and therefore `fused_init_helper_` will init them using `fvo_init_fn_`, when I think you want to init them using Wq and Wk using `qk_init_fn_`

vchiley (changes requested):
A few changes are needed

samhavens:
#210 adds (a lot of cool stuff including) a `model.gpt_block` config. I think maybe this PR should conform to that and use `gpt_block = 'post_ln'` instead of `norm_location = 'post_ln'`

examples/llm/src/models/param_init_fns.py
samhavens: I don't know how to change `fused_init_helper_` in a forward compatible way. I currently just want to make it `fused_init_helper_(module, qk_init, fvo_init)`, which I know makes it less general... but I am not sure what type of generality we will want in the future?

vchiley:
I like the `gpt_block = 'post_ln'` convention instead of a `norm_location`.
The convention can then be generalized to other block styles as well (ie PaLM)

examples/llm/src/models/param_init_fns.py
vchiley: Since its a super custom init_fn_, just dont use the fused init helper and do it by hand

dakinggg:
Closing, because llm has moved to llm-foundry.

PR #234, opened 2023-03-13
Add prefix lm option to MosaicGPT

alextrott16:
This PR introduces support for a Prefix-LM version of MosaicGPT. 
It slightly refactors the triton FA implementation to add `TritonFlashPrefixAttention` and adds logic to handle a prefix-lm-specific input `prefix_mask` (which controls which portions of the token sequence use bidirectional (inside the prefix) or causal attention).
I'm also slightly refactoring some of the code to re-use a utility for inferring the value of the required `bidirectional_mask` batch input.

### DIFF
--- a/examples/llm/src/models/layers/attention.py
+++ b/examples/llm/src/models/layers/attention.py
@@ -82,10 +82,16 @@ class TorchCausalAttention(nn.Module):
         batch_size,
         heads,
         seq_len,
+        prefix_mask=None,
         key_padding_mask=None,
         alibi=False,
         dtype=None,
     ):
+
+        if prefix_mask is not None:
+            raise NotImplementedError(
+                'This attention implementation does not support use of `prefix_lm`.'
+            )
 
         # select seq_len subset of attn mask
         attn_mask = attn_mask[..., :seq_len, :seq_len]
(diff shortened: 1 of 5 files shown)


examples/llm/src/models/layers/attention.py
```diff
@@ -82,11 +82,17 @@ def generate_attn_mask(
...
         key_padding_mask=None,
         alibi=False,
         dtype=None,
     ):
 
+        if prefix_mask is not None:
+            raise NotImplementedError(
+                'This attention implementation does not support use of `prefix_lm`.'
+            )
+
```
vchiley, line 95: Is there anything that stops us from supporting prefix_mask for torch impl (non-causal)?

alextrott16: My laziness. I suppose I should add support for that.

alextrott16: @vchiley Actually, is it worth messing around with the torch implementation? It has a deprecation warning and is already missing support for a number of features we seem to be moving towards.

vchiley: Given the attn rewrite, all attn impls are merged into one cls.
The only version that cannot support bias is flash and an error gets raised if you pass an attn_bias to it.

vchiley:
converting to draft.
@alextrott16 has a new version incoming.

abhi-mosaic:
Closing in favor of #248

PR #235, opened 2023-03-13
Expand S2S range and make pad trimming controllable

alextrott16:
This PR adds a couple improvements to the Mixture-of-Denoisers collator used for things like span corruption or UL2.
- It adds functionality to the sequence corruption so that the extreme end of the allowable parameter range essentially creates Causal LM training examples.
- It adds an `allow_pad_trimming` argument to control whether the collator is allowed to trim excess padding (this can speed things up but isn't always desired because variable sequence lengths can mess with memory). This trimming is now off by default.


abhi-mosaic (approved):
LGTM! Sorry for the delay in reviewing.

PR #240, opened 2023-03-16
refactor attn layers

vchiley:
This PR  refactors all attn_impl to use one cls which calls diff attn fns (flash, triton, torch) internally.
This means the state dict of one attn_impl is guaranteed to be the same as the other (this is tested now).

Note: this PR uses unpacked attn variants (ie q, k, v are chunked out from qkv), since kv_caching will concatenated tokens to kv and therefore q will be diff seq len from k and v.

Test also now compare all variants to each other with clip, qkln, and alibi (when possible). This is like 55 really quick tests.

@dskhudia lmk if this works for the inference work

@dskhudia also noted that we'll want to move to [torch.nn.functional.scaled_dot_product_attention](https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html#torch.nn.functional.scaled_dot_product_attention). maybe now, maybe later.

@samhavens re ift. Note: the flash path will unpad the input and do the attn calculation unpadded; the other impl should do calc on the entire pad input but it should be only up to the longest seq in the batch (not the max seq len of the model)


examples/llm/src/models/layers/attention.py
```diff
@@ -12,377 +13,320 @@
...
-                causal=True,
-                need_weights=False)
-            return self.out_proj(rearrange(
-                context, 'b s h d -> b s (h d)')), attn_weights
+    if is_causal:
+        s = max(s_q, s_k)
+        causal_mask = attn_weight.new_ones(s, s, dtype=torch.bool)
+        causal_mask.tril_()
+        causal_mask.logical_not_()
+        causal_mask = causal_mask[-s_q:, -s_k:]
```
bmosaicml, line 63: Is this for when we use KV caching and retrieve those attn weights directly from cache?

bmosaicml (approved):
Great work Vitaliy this looks clean and ready for caching

vchiley:
Running a 125m (with clip 6) using flash, triton, torch in wandb proj: [attn_refactor](https://wandb.ai/mosaic-ml/attn_refactor). MFU is about where it should be.
wandb proj: [attn_refactor](https://wandb.ai/mosaic-ml/attn_refactor) also has a [7b run with triton](https://wandb.ai/mosaic-ml/attn_refactor/runs/2legub78?workspace=user-vchiley) to verify MFU is where it should be (using 16 GPUs) (its at > 43% MFU; checks out)

examples/llm/src/models/layers/attention.py
```diff
@@ -12,377 +13,343 @@
...
+    v = rearrange(value, 'b s (h d) -> b h s d', h=n_heads)
+
+    b, _, s_q, d = q.shape
+    s_k = k.size(-1)
+
+    if softmax_scale is None:
+        softmax_scale = 1 / math.sqrt(d)
+
+    attn_weight = q.matmul(k)
+    attn_weight *= softmax_scale
```
abhi-mosaic, line 43:
```suggestion
    attn_weight = q.matmul(k) * softmax_scale
```

I have a deep fear of using += or *= style operations on torch tensors. Long ago there were autograd issues with doing so.

examples/llm/src/models/layers/attention.py
```diff
@@ -12,377 +13,343 @@
...
         self.d_model = cfg.d_model
         self.n_heads = cfg.n_heads
+        self.softmax_scale = cfg.get('softmax_scale')
+        if self.softmax_scale is None:
+            self.softmax_scale = 1 / math.sqrt(self.d_model / self.n_heads)
+        self.attn_dropout_p = cfg.get('attn_pdrop')
+
+        self.Wqkv = nn.Linear(self.d_model,
+                              3 * self.d_model,
+                              bias=True,
```
abhi-mosaic, line 228: should this be hardcoded true? Just want to make sure this is what we have done before

examples/llm/src/models/layers/attention.py
```diff
@@ -12,377 +13,343 @@
...
+                              device=device)
+        # for param init fn; enables shape based init of fused layers
+        fuse_splits = (cfg.d_model, 2 * cfg.d_model)
```
abhi-mosaic, line 231: Isn't this `3*cfg.d_model` since we have fused the Q,K,V?

examples/llm/src/models/layers/attention.py
```diff
@@ -12,377 +13,350 @@
...
+            warnings.warn(
+                'While `attn_impl: triton` can be faster than `attn_impl: flash` '
+                'it uses more memory. When training larger models this can trigger '
+                'alloc retries which hurts performance. If encountered, we recommend '
+                'using `attn_impl: flash`.')
+        elif self.attn_impl == 'torch':
+            self.attn_fn = scaled_multihead_dot_product_attention
+            warnings.warn(
+                DeprecationWarning(
+                    'Using `attn_impl: torch` is deprecated; recommened using `attn_impl: flash`.'
```
dskhudia, line 254: We are no longer deprecating it.

examples/llm/src/models/layers/attention.py
```diff
@@ -12,377 +13,343 @@
...
+        fuse_splits = (cfg.d_model, 2 * cfg.d_model)
+        self.Wqkv._fused = (0, fuse_splits)  # type: ignore
+
+        if self.attn_qk_ln:
+            layernorm_class = nn.LayerNorm if not cfg.get(
+                'low_precision_layernorm', False) else LPLayerNorm
+            self.q_ln = layernorm_class(self.d_model, device=device)
+            self.k_ln = layernorm_class(self.d_model, device=device)
+
+        self.is_causal = True  # causal attn impl
```
abhi-mosaic, line 240: I suggest deleting this and moving `is_causal: bool` to the `MultiheadAttention.forward()` args, because for PrefixLM (needed soon both internal and customer facing), we would like to set this to `False`, right?

examples/llm/src/models/layers/attention.py
dskhudia: However, you can add a warning that it can be slow.

dskhudia:
For inference, ideally we should have export test for onnx/torchscript but can be added in a followup PR.

examples/llm/src/models/layers/attention.py
```diff
@@ -12,377 +13,343 @@
...
+                needs_weights=False):
+        qkv = self.Wqkv(x)
+        if self.clip_qkv:
+            qkv.clamp_(min=-self.clip_qkv, max=self.clip_qkv)
+        if self.attn_qk_ln:
+            # Applying layernorm to qk
+            dtype = qkv.dtype
+            q, k, v = qkv.split(self.d_model, dim=-1)
+            q = self.q_ln(q).to(dtype)
+            k = self.k_ln(k).to(dtype)
```
abhi-mosaic, line 278: are these `.to(dtype)` calls necessary? Is this to make sure that when using `nn.LayerNorm`, which upcasts to `fp32`, we immediately downcast the activations back to 16-bit?

abhi-mosaic (approved):
This looks awesome @vchiley ! I added a few comments but they're minor.

dskhudia:
Looks great. Much cleaner now. 

> also noted that we'll want to move to [torch.nn.functional.scaled_dot_product_attention](https://pytorch.org/docs/stable/generated/torch.nn.functional.scaled_dot_product_attention.html#torch.nn.functional.scaled_dot_product_attention).

^ can be done later once we move to 2.0.

examples/llm/src/models/layers/attention.py
```diff
@@ -12,377 +13,343 @@
...
+                ))
+        else:
+            raise ValueError(f"{cfg.get('attn_impl')=} is an invalid setting.")
 
-        if self.attn_qk_ln or self.clip_qkv:
-            self.Wqkv = nn.Linear(self.d_model,
-                                  3 * self.d_model,
+        self.out_proj = nn.Linear(self.d_model,
+                                  self.d_model,
                                   bias=True,
```
abhi-mosaic, line 261: Same comment here re. hardcoded `bias=True`

examples/llm/src/models/layers/attention.py
vchiley: this defines the splits
the first split happens at `cfg.d_model` (end of first matrix and start of second matrix), the second split happens at `2 * cfg.d_model` (end of second matrix and start of third matrix). There are 3 matrices so no more splits.

dskhudia (approved):

examples/llm/src/models/layers/attention.py
vchiley: I added `self.is_causal` in prep for `PrefixLM`. Wasn't sure where to put it, but I'll move it to `MultiheadAttention.forward()` as suggested

examples/llm/src/models/layers/attention.py
vchiley: yes

examples/llm/src/models/layers/attention.py
vchiley: It already defaults to True so I'm not sure why i have this. I'll rm it cuz why have extra lines.

examples/llm/src/models/layers/attention.py
vchiley: My way: attn_weight is generated (`attn_weight = q.matmul(k)`), attn_weight is in-place modified (`attn_weight *= softmax_scale`)

Your way: `q.matmul(k)` instantiates `attn_weight`, ` * softmax_scale` is NOT an in-place mul of `attn_weight` so another set of attn_weights are generated...

I'll do:
```suggestion
    attn_weight = q.matmul(k)
    attn_weight.mul_(softmax_scale)
```
to avoid using += or *= style operations on torch tensors.

dskhudia: Actually let us not make them in-place. In-place operations don't interact nicely with other components including torch.compile. Torch.compile will anyway fuse these so keeping them separate is fine.

vchiley: I'll nuke a lot of in-place ops then

PR #241, opened 2023-03-16
Fdiff updt

vchiley:
Updt fdiff to only plot fdiff if it has seen metric before
Still works with train and eval metrics but should be more robust to adding other metrics (eg ICL) when they have unique logging intervals.

Issue: https://mosaicml.atlassian.net/browse/RESEARCH-551

### DIFF
--- a/examples/common/fdiff.py
+++ b/examples/common/fdiff.py
@@ -7,8 +7,6 @@
 """Monitor rate of change of loss."""
 from __future__ import annotations
 
-from typing import Any, Dict
-
 from composer.core import Callback, State
 from composer.loggers import Logger
 
@@ -16,8 +14,8 @@ from composer.loggers import Logger
 class FDiffMetrics(Callback):
     """Rate of chage of metrics.
 
-    tracks and plots the rate of change of metrics effectively taking the numerical
-    derivative of the metrics
+    tracks and plots the rate of change of metrics effectively taking the
+    numerical derivative of the metrics
     """
 
     def __init__(self, diff_train_metrics=True, diff_eval_metrics=True):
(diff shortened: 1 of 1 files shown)


examples/common/fdiff.py
```diff
@@ -7,20 +7,18 @@
...
     """Rate of chage of metrics.
 
-    tracks and plots the rate of change of metrics effectively taking the numerical
-    derivative of the metrics
+    tracks and plots the rate of change of metrics effectively taking the
+    numerical derivative of the metrics
     """
 
-    def __init__(self, diff_train_metrics=True, diff_eval_metrics=True):
+    def __init__(self, diff_train_metrics=False, diff_eval_metrics=True):
```
dakinggg, line 21: unintentional default change im guessing?

examples/common/fdiff.py
```diff
@@ -38,26 +36,31 @@ def batch_end(self, state: State, logger: Logger):
...
+                if mkey in self.eval_prev_metric.keys():
                     logger.log_metrics({
-                        f'metrics/eval/{k}_fdiff':
+                        f'{mkey}_fdiff':
                             state.eval_metric_values[k].item() -
-                            self.eval_prev_metric[k]
+                            self.eval_prev_metric[mkey]
                     })
 
             for k in state.eval_metric_values.keys():
```
dakinggg, line 63: should this loop be over `metrics` instead of `state.eval_metric_values`? I think both will work (cause it will just overwrite with the same value a bunch of times), but the second is more correct?

dakinggg (approved):
LGTM, would appreciate pasted in wandb plots from a manual test

examples/common/fdiff.py
vchiley: diff_train_metrics are super high variance making it almost impossible to get a real signal from fdiff.
It would effectively require a moving window or avg, but thats a bit out the scope of this pr.

examples/common/fdiff.py
vchiley: good catch.

PR #247, opened 2023-03-23
Fix non assignment bug

alextrott16:
Hot fixes issue in Mosaic GPT in latest updates that doesn't update the attention bias. Fixes alibi.


vchiley (approved):
❤️

PR #248, opened 2023-03-23
Add prefix_lm support to MosaicGPT

alextrott16:
Adds `prefix_lm` config option that tells the model to expect a `prefix_mask` input. Tokens where `prefix_mask==1` can attend to one another bidirectionally.

Confirmed correct behavior manually for `torch` (w/ and w/o `alibi`), using `text_denoising` data.

Still missing unit tests.


examples/llm/src/models/layers/attention.py
```diff
@@ -366,20 +368,24 @@ def forward(self,
...
+            return (1, n_heads, seq_len, seq_len)
         return None
     elif attn_impl == 'torch':
         if alibi:
             if not causal:
                 return (1, n_heads, seq_len, seq_len)
             return (1, n_heads, 1, seq_len)
+        elif prefix_lm:
+            return (1, n_heads, seq_len, seq_len)
         return None
```
abhi-mosaic, line 389: Squinting hard here... are both these blocks the same? Can we just do `elif attn_impl in ['triton', 'torch']`

(this is also from a previous PR so not your fault)

abhi-mosaic (approved):
This looks super clean! Approving to unblock and gonna rely on your testing @alextrott16

examples/llm/src/models/mosaic_gpt.py
```diff
@@ -132,11 +136,38 @@ def _attn_bias(self, device, dtype):
 
         return self.attn_bias
 
+    def _apply_prefix_mask(self, attn_bias: torch.Tensor,
+                           prefix_mask: torch.Tensor):
+        assert attn_bias.shape[-1] == self.config.max_seq_len
```
vchiley, line 141: raise not assert.

also shouldn't it be `attn_bias.shape[-1] == self.config.max_seq_len and attn_bias.shape[-2] == self.config.max_seq_len`

examples/llm/src/models/mosaic_gpt.py
```diff
@@ -132,11 +136,38 @@ def _attn_bias(self, device, dtype):
...
+            )
+
+        # select seq_len subset of attn mask
+        attn_bias = attn_bias[..., :seq_len, :seq_len]
+
+        # Mix the causal max and the bidirectional mask to get the full
+        # allowable attention (i.e. full = not accounting for padding yet)
+        causal = torch.tril(
+            torch.ones((seq_len, seq_len),
+                       dtype=torch.uint8,
```
vchiley, line 155:
```suggestion
                       dtype=torch.bool,
```

examples/llm/tests/test_flash_triton_torch.py
```diff
@@ -61,16 +61,17 @@ def gen_bias(attn_impl):
...
         if bs is not None:
             attn_bias = torch.zeros(*bs, device=device)
-            attention.attn_bias(attn_impl,
-                                attn_bias,
-                                cfg.n_heads,
-                                s,
-                                causal=causal,
-                                alibi=alibi,
-                                alibi_bias_max=8)
+            attn_bias = attention.attn_bias(attn_impl,
```
vchiley, line 68: TY

vchiley (approved):
Generally looks good. Left a few comments.

examples/llm/src/models/mosaic_gpt.py
```diff
@@ -213,6 +249,11 @@ def forward(
 
         attn_bias = self._attn_bias(device=x.device, dtype=x.dtype)
 
+        if self.prefix_lm:
+            assert isinstance(attn_bias, torch.Tensor)  # pyright
+            assert isinstance(prefix_mask, torch.Tensor)  # pyright
```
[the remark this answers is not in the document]
alextrott16: TBH these are just to get pyright to get off its high horse. The rest of the code seems to allow for these pyright-pacifying asserts (even just 5 lines above these ones, for example). So I figured it was OK.

Do you think raising an error makes more sense here?

examples/llm/src/models/mosaic_gpt.py
alextrott16: Since this method should only trigger for a prefix_lm, yes. Good catch!

examples/llm/src/models/mosaic_gpt.py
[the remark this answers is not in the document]
dakinggg: if its just for pyright, assert is all good.

examples/llm/src/models/mosaic_gpt.py
```diff
@@ -162,6 +193,11 @@ def forward(
             raise NotImplementedError(
                 'MosaicGPT does not support training with left padding.')
 
+        if self.prefix_lm and prefix_mask is None:
+            raise ValueError(
+                'prefix_mask is a required argument when is configured with prefix_lm=True.'
```
dakinggg, line 198:
```suggestion
                'prefix_mask is a required argument when MosaicGPT is configured with prefix_lm=True.'
```

examples/llm/src/models/mosaic_gpt.py
```diff
@@ -315,11 +368,17 @@ def get_targets(self, batch):
         return targets
 
     def forward(self, batch):
+        if self.model.prefix_lm:
```
dakinggg, line 371: The answer may be no (and I haven't been able to think of one), but is there anywhere we can put this logic other than here? Trying to avoid custom logic in `ComposerMosaicGPT` if possible.

examples/llm/src/models/mosaic_gpt.py
```diff
@@ -279,9 +320,21 @@ def prepare_inputs_for_generation(self,
         if past_key_values is not None:
             input_ids = input_ids[:, -1].unsqueeze(-1)
 
+        if self.prefix_lm:
+            # Leverage a convenience of sequential generation!
+            prefix_mask = torch.ones_like(attention_mask)
+            # This requires that we're using the cache
+            if kwargs.get('use_cache') == False:
```
dakinggg, line 327: why is it that cache is required for this?

examples/llm/src/models/mosaic_gpt.py
alextrott16: This same logic is actually used by our HF prefix LM. It's mostly a workaround to support ICL eval for prefix LMs.
We can remove these logics once the rest of our upstream stack embraces prefix masks... Someday...

dakinggg: That is the answer I expected, sounds good :)

examples/llm/src/models/mosaic_gpt.py
alextrott16: In order to use the trick that prefix_mask=1 everywhere, it needs to be the case that previous tokens (i.e. tokens that were supplied as the input or already generated) do not have their key/value embeddings updated. Otherwise, this trick would cause those embeddings to change each step, which would be inconsistent with how the model was trained.

dakinggg: ahh got it, thanks!

PR #250, opened 2023-03-23
upgrade transformers and add unit test for generate with cache

dakinggg:
This PR upgrades the transformers version because the `past` arg was renamed to `past_key_values`. When I was doing all my testing, I was running with `4.26.1` which is not the version specified in requirements right now.

TODO:

- [x] start a run with 4.26.1 and check that the output loss curve is unchanged


alextrott16 (approved):
Rocketship emoji

PR #252, opened 2023-03-24
Use Composer HuggingFaceModel base class

dakinggg:
This PR makes `ComposerMosaicGPT` a `HuggingFaceModel`, to give us extra functionality, and reduce the duplicate code necessary here.

TODO:
- [x] test mosaicgpt before and after
- [x] test gpt-neo before and after

Note: the gptneo runs diverge, i probably used some bad settings, the important thing is that they are the same before and after
Screen Shot 2023-03-24 at 10 43 46 AM

- [x] test inference script before and after (not doing because it is not working on main at the moment)
- [x] test evaluate script before and after
gpt-neo-2.7b, same lambada accuracy before and after


abhi-mosaic (approved):
Thanks for consolidating the codepaths! Excited to get the extra features from Compser's `HuggingFaceModel` wrapper class.

PR #255, opened 2023-03-25
Add torch 2.0 based tensor parallel support

dskhudia:
Add torch2.0 based tensor parallel support. By default, tensor parallel for MLPs is off. 

It doesn't affect the code when using torch < 2.0.0. if model.tensor_parallel_mlp=true with torch < 2.0.0, the following error is printed out. 

`Tensor Parallel with torch version < 2.0.0 is not implemented. Please set tensor_parallel_mlp to false.` 

~~Will add runs with and without this with torch 2.0.~~
- [x] [wandb link](https://wandb.ai/mosaic-ml/daya-tensor-parallel/reports/Compare-TP-vs-No_TP--VmlldzozODg4ODE3) PairwiseParallel

Tensor parallel based on  PairwiseParallel  requires that inputs to the tensor parallel ranks be the same. There is another version of it called PairwiseSeqParallel that all gathers inputs for the tensor parallel ranks and somehow that is not part of the 2.0 release. ~~I plan to add it later.~~ Added PairwiseSeqParallel. 

Reasons to check this in:

1. This is 0 => 1 and we can build upon and improve it
2. Others can also debug the performance issues.

### DIFF
--- a/examples/llm/src/models/mosaic_gpt/configuration_mosaic_gpt.py
+++ b/examples/llm/src/models/mosaic_gpt/configuration_mosaic_gpt.py
@@ -7,6 +7,8 @@ from typing import Optional, Tuple, Union
 
 from transformers import PretrainedConfig
 
+from examples.llm.src.models.utils import is_torch_2_or_higher
+
 
 class MosaicGPTConfig(PretrainedConfig):
     model_type = 'mosaic_gpt'
@@ -45,6 +47,7 @@ class MosaicGPTConfig(PretrainedConfig):
         embedding_fraction: float = 1.0,
         low_precision_layernorm: bool = False,
         use_cache: bool = True,
+        tensor_parallel_mlp: bool = False,
         **kwargs,
     ):
         """The MosaicGPT configuration class.
@@ -88,6 +91,7 @@ class MosaicGPTConfig(PretrainedConfig):
             embedding_fraction (float): The fraction to scale the gradients of the embedding layer by.
             low_precision_layernorm (bool): Whether to use low precision layer normalization.
             use_cache (bool): Whether or not the model should return the last key/values attentions
+            tensor_parallel_mlp (bool): Whether or not to tensor parallelize the MLPs
         """
         self.d_model = d_model
         self.n_heads = n_heads
@@ -121,6 +125,7 @@ class MosaicGPTConfig(PretrainedConfig):
         self.embedding_fraction = embedding_fraction
         self.low_precision_layernorm = low_precision_layernorm
         self.use_cache = use_cache
+        self.tensor_parallel_mlp = tensor_parallel_mlp
         if 'name' in kwargs:
             del kwargs['name']
         if 'loss_fn' in kwargs:
@@ -153,4 +158,8 @@ class MosaicGPTConfig(PretrainedConfig):
                       str) and self.logit_scale != 'inv_sqrt_d_model':
             raise ValueError(
                 f"{self.logit_scale=} is not recognized as an option; use numeric value or 'inv_sqrt_d_model'."
+            )
+        if not is_torch_2_or_higher() and self.tensor_parallel_mlp:
+            raise NotImplementedError(
+                'Tensor Parallel with torch version < 2.0.0 is not implemented. Please set tensor_parallel_mlp to false.'
             )
--- a/examples/llm/src/models/mosaic_gpt/mosaic_gpt.py
+++ b/examples/llm/src/models/mosaic_gpt/mosaic_gpt.py
@@ -19,6 +19,7 @@ from composer.metrics import (InContextLearningLMAccuracy,
                               InContextLearningMultipleChoiceAccuracy)
 from composer.metrics.nlp import LanguageCrossEntropy, LanguagePerplexity
 from composer.models import HuggingFaceModel
+from composer.utils import dist, get_device
 from omegaconf import DictConfig
 from omegaconf import OmegaConf as om
 from transformers import AutoTokenizer, PreTrainedModel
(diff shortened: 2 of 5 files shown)


examples/llm/src/models/mosaic_gpt/configuration_mosaic_gpt.py
```diff
@@ -45,6 +46,9 @@ def __init__(
         embedding_fraction: float = 1.0,
         low_precision_layernorm: bool = False,
         use_cache: bool = True,
+        tensor_parallel_mlp: bool = False,
+        world_size: int = 1,
+        local_world_size: int = 1,
```
abhi-mosaic, line 51: Is there any reason these values cannot be calculated dynamically in the model or layer init?

If I share a model config with someone else, I don't think they need to know the dist information I used at training time, and in fact they may change it if they want to train/inference on a different set of machines. 

Tagging @vchiley because we have a similar issue with MoE layer init and I want to make sure we do the same thing in both places.

examples/llm/src/models/mosaic_gpt/mosaic_gpt.py
```diff
@@ -64,14 +70,19 @@ def __init__(self, config: MosaicGPTConfig):
...
-                    for _ in range(config.n_layers)
-                ])
-        })
+        layers = nn.ModuleList()
+        for _ in range(config.n_layers):
+            block = gpt_blocks.GPTBlock(device=config.init_device, **config.to_dict())
+            if config.tensor_parallel_mlp and is_torch_2_or_higher() and enable_2d_with_fsdp():
+                device_type = 'cuda' if get_device(None).name == 'gpu' else 'cpu'
+                node_count = config.world_size // config.local_world_size
+                twod_mesh = DeviceMesh(device_type=device_type, mesh=torch.arange(0, config.world_size).view(node_count, -1))
```
abhi-mosaic, line 79: I wonder if we should be creating `DeviceMesh` just once, and then passing to each block?

dakinggg:
sorry if i missed part of the conversation somewhere, but whats up with the difference in loss curves between with tp and without tp?

examples/llm/src/models/mosaic_gpt/configuration_mosaic_gpt.py
dskhudia: I didn't want to add dist dependency on init. It may or may not be initialized at init. For example, as you mentioned, during inference it may not be available and having dist dependency breaks model init.

examples/llm/src/models/mosaic_gpt/mosaic_gpt.py
dskhudia: Let me rearrange the code. It will take care of this.

dskhudia:
> sorry if i missed part of the conversation somewhere, but whats up with the difference in loss curves between with tp and without tp?

The current curves (with pairwiseparallel) are supposed to be different as this style requires inputs to the TP ranks the same. We currently don't do same inputs for the TP ranks. PairWiseSeqParallel is gathers inputs from the tp ranks so it should show similar curves (experiments are currently running). Will post links soon.

examples/llm/src/models/mosaic_gpt/configuration_mosaic_gpt.py
dakinggg: composer dist should just return a default value (1) if dist is not initialized. does that make it ok to just do inside the model instead of as a config var?

dskhudia: Yes. That should work.

dskhudia: Removed these config params.

examples/llm/src/models/mosaic_gpt/configuration_mosaic_gpt.py
```diff
@@ -88,6 +92,9 @@ def __init__(
             embedding_fraction (float): The fraction to scale the gradients of the embedding layer by.
             low_precision_layernorm (bool): Whether to use low precision layer normalization.
             use_cache (bool): Whether or not the model should return the last key/values attentions
+            tensor_parallel_mlp (bool): Whether or not to tensor parallelize the MLPs
+            world_size (int): world size for training. Used in tensor parallelizing the mlp block within a node.
```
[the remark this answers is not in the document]
dskhudia: Removed both of these options. Now we use dist apis directly during model init.

examples/llm/src/models/utils/tensor_parallel.py
```diff
@@ -0,0 +1,122 @@
...
+                                               make_input_shard_1d,
+                                               make_output_shard_1d)
+from torch.distributed.tensor.parallel._utils import (_prepare_input_validate,
+                                                      _prepare_output_validate)
+
+
+def backward(ctx, grad_output: 'dtensor.DTensor'):  # type: ignore[override]
+    previous_placement = ctx.previous_placement
+    previous_device_mesh = ctx.previous_device_mesh
+    # When we run backward pass of redistribute (i.e. manual redistribute from
```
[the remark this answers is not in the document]
dskhudia: This is just a comment for the for loop rather than function docstring.

examples/llm/src/models/mosaic_gpt/mosaic_gpt.py
```diff
@@ -72,6 +81,22 @@ def __init__(self, config: MosaicGPTConfig):
                     for _ in range(config.n_layers)
                 ])
         })
+        if config.tensor_parallel_mlp and is_torch_2_or_higher(
+        ) and enable_2d_with_fsdp():
```
dakinggg, line 85: looks like right now if `enable_2d_with_fsdp` fails it will silently not apply. Could we break that out into a separate exception? (I don't know the context in which that would fail, so feel free to disregard if this doesn't make sense)

examples/llm/src/models/utils/tensor_parallel.py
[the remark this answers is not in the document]
dakinggg: this file is also copied from pytorch, so probably better not to modify unless we need to

dakinggg (approved):
Would be great if you could add some basic tests (even though they won't run until we upgrade to torch 2.0), but otherwise looks good to me!

dskhudia:
> Would be great if you could add some basic tests
Will add and gate them with torch_2_0 check.

examples/llm/src/models/mosaic_gpt/mosaic_gpt.py
dskhudia: will do and raise an exception.

examples/llm/src/models/utils/tensor_parallel.py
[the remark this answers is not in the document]
dskhudia: It's copied from multiple different files. I still need to make formatting gods happy so have to modify this. 😄

dakinggg:
we haven't set them up in CI yet (sorry), but you can use @pytest.mark.gpu, and then when we do they will run. for now people are just running the gpu tests locally as needed.

dakinggg:
Closing, because this PR has moved to llm-foundry.

PR #259, opened 2023-03-28
fix bias

vchiley:
The fundamental fix is adding: `attn_bias = attn_bias[:, :, :, -s_k:]`. Also moved stuff around a bit.

Parametrize a bunch of model tests with alibi.


dakinggg (approved):
lgtm

PR #261, opened 2023-03-28
Use HF generate for inference

dskhudia:
Now we use HF generate interface for inference.

### DIFF
--- a/examples/llm/inference/inference.py
+++ b/examples/llm/inference/inference.py
@@ -1,87 +1,14 @@
 # Copyright 2022 MosaicML Examples authors
 # SPDX-License-Identifier: Apache-2.0
+import sys
 import warnings
-from typing import List
 
 import torch
 from composer.core import get_precision_context
+from composer.utils import get_device
 from omegaconf import OmegaConf as om
 
 from examples.llm.src import COMPOSER_MODEL_REGISTRY
-
-
-def sample_top_p(probs, p):
-    probs_sort, probs_idx = torch.sort(probs, dim=-1, descending=True)
-    probs_sum = torch.cumsum(probs_sort, dim=-1)
-    mask = probs_sum - probs_sort > p
-    probs_sort[mask] = 0.0
-    probs_sort.div_(probs_sort.sum(dim=-1, keepdim=True))
-    next_token = torch.multinomial(probs_sort, num_samples=1)
-    next_token = torch.gather(probs_idx, -1, next_token)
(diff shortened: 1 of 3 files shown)


examples/llm/inference/inference.py
```diff
@@ -111,9 +38,47 @@ def get_mosaicgpt_inference_model(checkpoint_yaml_path: str, tokenizer):
 
     model.load_state_dict(checkpoint['state']['model'], strict=True)
 
-    model.cuda()
-    model.eval()
+    model.tokenizer.pad_token_id = 50256
```
dakinggg, line 41: maybe modify this to only set it if the tokenizer doesn't already have a pad token id? we may use a different tokenizer in the future that already has the pad token set.

dakinggg (approved):
LGTM, thanks Daya!

examples/llm/inference/inference.py
dskhudia: Fixed. Is there anything specific about 50256 or you just picked the last token?

dskhudia: Just wondering how vocab_size would affect this.

dakinggg: 50256 is the eos token in the gpt tokenizer. thats it

dakinggg: maybe just set it to tokenizer.eos_token_id instead of 50256, and emit a warning about doing that. cause if the vocab is smaller than 50256, it won't work

examples/llm/inference/inference.py
```diff
@@ -109,11 +36,55 @@ def get_mosaicgpt_inference_model(checkpoint_yaml_path: str, tokenizer):
...
+    if 'state' in checkpoint.keys():
+        # it's a full training checkpoint
+        model.load_state_dict(checkpoint['state']['model'], strict=True)
+    else:
+        # it's a weights-only checkpoint
+        model.load_state_dict(checkpoint, strict=True)
 
-    model.cuda()
-    model.eval()
+    if not model.tokenizer.pad_token_id:
```
dakinggg, line 46:
```suggestion
    if model.tokenizer.pad_token_id is not None:
```
in case the pad token id is 0

dskhudia: Done

examples/llm/inference/inference.py
dskhudia: Done

PR #265, opened 2023-03-30
Add a callback that logs generations to wandb at eval end

dakinggg:


examples/common/generate_callback.py
```diff
@@ -0,0 +1,67 @@
...
+                            tokenized_input['input_ids'].shape[1]:]
+                        output_text = tokenizer.decode(output_tokens,
+                                                       skip_special_tokens=True)
+
+                        rows.append([prompt, output_text])
+
+                    text_table = wandb.Table(data=rows,
+                                             columns=['prompt', 'generation'])
+                    wandb_artifact.add(text_table, 'predictions')
+                    wandb.log_artifact(wandb_artifact)
```
abhi-mosaic, line 65: QQ, is this preferable to just logging the table via `wandb.log({"generate_samples" : text_table})`

abhi-mosaic: Artifacts always seem to add complexity... maybe we can also just log both if the reasoning is that that the Artifacts are easier to download?

abhi-mosaic:
LOVE IT!

examples/common/generate_callback.py
dakinggg: I started out with what you had...wandb does _not_ make it easy to view the evolution of this over the course of training. As far as I can tell the two options are

1. Log as table. The table in the UI will get overwritten  each time you write a new one, but it will show up in the main chart view (with an absurd amount of whitespace)
2. Log as artifact. The table will not show up in the main chart view, but if you click through like five menus, you have the ability to view all the generations over the course of training (in a pretty rough UI).

Do you have a different suggestions based on your wandb knowledge? Entirely possible I missed something.

vchiley: Should `self.wandb_artifact = wandb.Artifact(*args)` instantiation be done durring callback init?

dakinggg: turns out no. once you send an artifact off to wandb, that object cannot be used anymore.
`ValueError: Can't add to finalized artifact.`

dakinggg: Discussed offline, going to log both as table and artifact to get the best of both worlds. its still a pretty rough UI

examples/common/generate_callback.py
```diff
@@ -0,0 +1,92 @@
...
+        7. Now, when you hover over other versions, there will be a "compare" button, which will allow you to compare the currently
+            selected version to the version you add via compare.
+
+        Args:
+            prompts (List[str]): The list of prompts you would like to produce generations for
+            kwargs: All kwargs well be passed along to the call to generate. This is for things like `do_sample`, `top_p`, etc
+        """
+        self.prompts = prompts
+        self.generate_kwargs = kwargs
+
```
vchiley, line 38:
```suggestion
        self.wandb_logger = None

```

examples/common/generate_callback.py
```diff
@@ -0,0 +1,92 @@
...
+    def init(self, state: State, logger: Logger):
+        if dist.get_global_rank() == 0:
+            for destination in ensure_tuple(logger.destinations):
+                if isinstance(destination, WandBLogger):
```
vchiley, line 42: do we care about [other loggers](https://docs.mosaicml.com/en/latest/trainer/logging.html#available-loggers)?

vchiley: or is this a problem for when the callback is upstreamed to composer?

dakinggg: im gonna say no; this is a wandb specific implementation. can be adapted in the future if desired.

dakinggg: basically this callback is a no-op if you're not using wandb. I think that is ok for now. if/when upstreamed i can think about supporting other loggers as makes sense.

vchiley (approved):

PR #266, opened 2023-03-30
Support only-within-sequence attention for MosaicGPT

alextrott16:
Makes it possible to have attention restricted to tokens within the same source sequence when using pre-concatenated text dataloading.

- Adds `eos_token_id` and `bos_token_id` flags for the `text` dataloader, which instructs it to add `"sequence_id"` to the batch using one of those tokens as the separator.
- Adds `attn_uses_sequence_id` flag to MosaicGPT config/model and `sequence_id` to the forward arguments.
- Modifies the attention bias construction to restrict attention based on `sequence_id` if the model config is set to use that input.

125m comparison: https://wandb.ai/mosaic-ml/v004-125m
3b comparison: https://wandb.ai/mosaic-ml/v004-3b/table?workspace=user-alextrott (for profiling; not a full convergence run)

### DIFF
--- a/examples/common/text_data.py
+++ b/examples/common/text_data.py
@@ -5,7 +5,7 @@
 
 import os
 from itertools import islice
-from typing import Any, Dict, Optional
+from typing import Any, Callable, Dict, List, Optional
 
 import numpy as np
 import torch
@@ -137,13 +137,65 @@ class StreamingTextDataset(StreamingDataset):
         return token_sample
 
 
+class ConcatenatedSequenceCollatorWrapper:
+    """Collator wrapper to add sequence_id to batch."""
+
+    def __init__(self,
+                 base_collator: Callable,
+                 eos_token_id: Optional[int] = None,
+                 bos_token_id: Optional[int] = None):
+        self.base_collator = base_collator
+        if (eos_token_id is None) and (bos_token_id is None):
+            raise ValueError(
+                'Must supply a value for either eos_token_id or bos_token_id, but got None for both.'
+            )
+        if (eos_token_id is not None) and (bos_token_id is not None):
+            raise ValueError(
+                'Must supply a value for either eos_token_id or bos_token_id, but both were given.'
+            )
+        if eos_token_id is None:
+            self.split_token_id = bos_token_id
+            self.eos_mode = False
+        else:
+            self.split_token_id = eos_token_id
+            self.eos_mode = True
+        assert self.split_token_id is not None
+
+    def __call__(self, examples: List[Any]) -> Dict[str, torch.Tensor]:
+        batch = self.base_collator(examples)
+        if self.eos_mode:
+            return self.sequence_id_from_eos(batch, self.split_token_id)
+        else:
+            return self.sequence_id_from_bos(batch, self.split_token_id)
+
+    @staticmethod
+    def sequence_id_from_eos(batch: Dict[str, torch.Tensor],
+                             split_token_id: int) -> Dict[str, torch.Tensor]:
+        is_eos = torch.eq(batch['input_ids'], split_token_id)
+        cumulative_eos = torch.cumsum(is_eos,
+                                      dim=1).to(batch['input_ids'].dtype)
+        left_zeros = cumulative_eos.new_zeros((cumulative_eos.shape[0], 1))
+        batch['sequence_id'] = torch.cat([left_zeros, cumulative_eos[:, :-1]],
+                                         dim=1)
+        return batch
+
+    @staticmethod
+    def sequence_id_from_bos(batch: Dict[str, torch.Tensor],
+                             split_token_id: int) -> Dict[str, torch.Tensor]:
+        is_bos = torch.eq(batch['input_ids'], split_token_id)
+        cumulative_bos = torch.cumsum(is_bos,
+                                      dim=1).to(batch['input_ids'].dtype)
+        batch['sequence_id'] = cumulative_bos
+        return batch
+
+
 def build_text_dataloader(cfg: DictConfig, device_batch_size: int):
     assert cfg.name == 'text', f'Tried to build text dataloader with cfg.name={cfg.name}'
     if cfg.dataset.get('group_method', None) is not None:
         raise NotImplementedError(
             'group_method is deprecated and has been removed.\nTo ' +
             'concatenate, use the --concat_tokens ' +
-            'argument when creating your MDS dataset with concat_c4.py')
+            'argument when creating your MDS dataset with convert_dataset.py')
     dataset = StreamingTextDataset(
         local=cfg.dataset.local,
         tokenizer_name=cfg.dataset.tokenizer_name,
@@ -165,6 +217,19 @@ def build_text_dataloader(cfg: DictConfig, device_batch_size: int):
         tokenizer=dataset.tokenizer,
         mlm=mlm_probability is not None,
         mlm_probability=mlm_probability)
+
+    if cfg.dataset.get('eos_token_id') is not None:
+        if cfg.dataset.get('bos_token_id') is not None:
+            raise ValueError(
+                'Cannot use *both* EOS and BOS tokens for detecting sequence boundaries. ' +\
+                'Please supply `eos_token_id` if sequences end with an EOS token, or use ' +\
+                '`bos_token_id` if sequences start with a BOS token.'
+            )
+        collate_fn = ConcatenatedSequenceCollatorWrapper(
+            base_collator=collate_fn, eos_token_id=cfg.dataset.eos_token_id)
+    elif cfg.dataset.get('bos_token_id') is not None:
+        collate_fn = ConcatenatedSequenceCollatorWrapper(
+            base_collator=collate_fn, bos_token_id=cfg.dataset.bos_token_id)
 
     return DataLoader(
         dataset,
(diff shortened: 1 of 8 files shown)


vchiley:
cc @jfrankle

alextrott16:
FYI, I think this PR has picked up some commits from the main branch that maybe weren't part of `release/v0.0.4` yet? There are some changed files that I didn't touch in any of my commits.

examples/common/text_data.py
```diff
@@ -137,13 +137,65 @@ def __getitem__(self, idx: int) -> Dict[str, Any]:
...
+        return batch
+
+    @staticmethod
+    def sequence_id_from_bos(batch: Dict[str, torch.Tensor],
+                             split_token_id: int) -> Dict[str, torch.Tensor]:
+        is_bos = torch.eq(batch['input_ids'], split_token_id)
+        cumulative_bos = torch.cumsum(is_bos,
+                                      dim=1).to(batch['input_ids'].dtype)
+        batch['sequence_id'] = cumulative_bos
+        return batch
```
vchiley, line 189: ❤️

examples/common/text_data.py
```diff
@@ -137,13 +137,65 @@ def __getitem__(self, idx: int) -> Dict[str, Any]:
...
+            raise ValueError(
+                'Must supply a value for either eos_token_id or bos_token_id, but both were given.'
+            )
+        if eos_token_id is None:
+            self.split_token_id = bos_token_id
+            self.eos_mode = False
+        else:
+            self.split_token_id = eos_token_id
+            self.eos_mode = True
+        assert self.split_token_id is not None
```
vchiley, line 162: informative value error instead of uninformative assert

examples/common/text_data.py
```diff
@@ -137,13 +137,65 @@ def __getitem__(self, idx: int) -> Dict[str, Any]:
...
+        if (eos_token_id is not None) and (bos_token_id is not None):
+            raise ValueError(
+                'Must supply a value for either eos_token_id or bos_token_id, but both were given.'
+            )
+        if eos_token_id is None:
+            self.split_token_id = bos_token_id
+            self.eos_mode = False
+        else:
+            self.split_token_id = eos_token_id
+            self.eos_mode = True
```
vchiley, line 161: Why use `self.split_token_id` and `self.eos_mode`?
Why not just set `self.eos_token_id` and `self.bos_token_id` then in call make the call based on those?

examples/common/text_data.py
```diff
@@ -137,13 +137,65 @@ def __getitem__(self, idx: int) -> Dict[str, Any]:
...
+    def __call__(self, examples: List[Any]) -> Dict[str, torch.Tensor]:
+        batch = self.base_collator(examples)
+        if self.eos_mode:
+            return self.sequence_id_from_eos(batch, self.split_token_id)
+        else:
+            return self.sequence_id_from_bos(batch, self.split_token_id)
+
+    @staticmethod
+    def sequence_id_from_eos(batch: Dict[str, torch.Tensor],
+                             split_token_id: int) -> Dict[str, torch.Tensor]:
```
vchiley, line 173: does this need to be a staticmethod?
if not, then you don't need to pass `split_token_id` you can just use `self.eos_token_id`
(assuming you use my previous suggestion)

same comment for `sequence_id_from_bos`

examples/common/text_data.py
```diff
@@ -166,6 +218,19 @@ def build_text_dataloader(cfg: DictConfig, device_batch_size: int):
         mlm=mlm_probability is not None,
         mlm_probability=mlm_probability)
 
+    if cfg.dataset.get('eos_token_id') is not None:
+        if cfg.dataset.get('bos_token_id') is not None:
+            raise ValueError(
+                'Cannot use *both* EOS and BOS tokens for detecting sequence boundaries. ' +\
+                'Please supply `eos_token_id` if sequences end with an EOS token, or use ' +\
+                '`bos_token_id` if sequences start with a BOS token.'
+            )
```
vchiley, line 227: isn't this already validated in `ConcatenatedSequenceCollatorWrapper`?

examples/llm/src/models/mosaic_gpt/mosaic_gpt.py
```diff
@@ -153,8 +156,14 @@ def _attn_bias(self,
             assert isinstance(prefix_mask, torch.Tensor)  # pyright
             attn_bias = self._apply_prefix_mask(attn_bias, prefix_mask)
 
+        # If using torch or triton, we incorporate sequence_id (if appropriate)
+        if self.attn_uses_sequence_id and sequence_id is not None:
+            assert isinstance(attn_bias, torch.Tensor)  # pyright
+            # assert isinstance(sequence_id, torch.Tensor)  # pyright
```
vchiley, line 162:
```suggestion
```

examples/llm/src/models/mosaic_gpt/mosaic_gpt.py
```diff
@@ -209,12 +218,34 @@ def _apply_prefix_mask(self, attn_bias: torch.Tensor,
...
+        # select seq_len subset of attn mask
+        attn_bias = attn_bias[..., :seq_len, :seq_len]
+
+        # Restrict attention to tokens that share the same value
+        # in sequence_id
+        cannot_attend = torch.logical_not(
+            torch.eq(sequence_id.view(-1, seq_len, 1),
+                     sequence_id.view(-1, 1, seq_len))).unsqueeze(1)
+        min_val = torch.finfo(attn_bias.dtype).min
+        attn_bias = attn_bias.masked_fill(cannot_attend, min_val)
```
vchiley, line 238: ❤️

examples/llm/src/models/mosaic_gpt/mosaic_gpt.py
```diff
@@ -242,6 +273,19 @@ def forward(
                 'prefix_mask is a required argument when MosaicGPT is configured with prefix_lm=True.'
             )
 
+        if self.training:
```
vchiley, line 276: [Abhi mention that our eval set is also preconcat and should use this.](https://mosaicml.slack.com/archives/C05000N6VD3/p1680129645540909)
(Sorry I know I'm the one who lead you astray on this one)

examples/llm/src/models/mosaic_gpt/mosaic_gpt.py
```diff
@@ -290,11 +334,13 @@ def forward(
             assert isinstance(self.transformer.emb_drop, nn.Module)  # pyright
             x = self.transformer.emb_drop(x_shrunk)
 
-        attn_bias, attention_mask = self._attn_bias(
-            device=x.device,
-            dtype=x.dtype,
-            attention_mask=attention_mask,
-            prefix_mask=prefix_mask)
+        with torch.no_grad():
```
vchiley, line 337: can we shove this into the `self._attn_bias` fn or better yet make it a decorator over that fn (see example [here](https://pytorch.org/docs/stable/generated/torch.no_grad.html) with decorator)

ex
```
    @torch.no_grad()
    def _attn_bias(self,
                   device,
                   dtype,
                   attention_mask: Optional[torch.ByteTensor] = None,
                   prefix_mask: Optional[torch.ByteTensor] = None):
                   prefix_mask: Optional[torch.ByteTensor] = None,
                   sequence_id: Optional[torch.LongTensor] = None):
```
when you define the fn above

examples/llm/src/models/mosaic_gpt/mosaic_gpt.py
```diff
@@ -364,6 +410,11 @@ def prepare_inputs_for_generation(self,
             raise NotImplementedError(
                 'MosaicGPT does not support generation with right padding.')
 
+        if self.attn_uses_sequence_id and self.training:
```
vchiley, line 413: would we ever be using the generate fn if model is in `self.training` mode?

examples/llm/tests/test_model.py
vchiley, line 1: i might be a good idea to add some prefixlm and seq_id tests here

vchiley (changes requested):
There are a few things to address + a few nit

examples/llm/src/models/mosaic_gpt/mosaic_gpt.py
alextrott16: I think we *have* to skip the input checking outside train mode. I'm fine with continuing to require `sequence_id` being supplied during training when `attn_uses_sequence_id==True`, but requiring it during eval mode is a huge burden.
Even though the input checking below this is skipped if the model is in eval mode, `sequence_id` will still be used if supplied (and the flag is turned on). That is, it'll handle our eval dataset correctly. So, I feel like this is a good compromise.

examples/llm/src/models/mosaic_gpt/mosaic_gpt.py
alextrott16: We might not end up doing so, but it seems like a user could. HF generate does not put the model into eval mode, as far as I can tell.

examples/common/text_data.py
alextrott16: It doesn't need to be, but being a staticmethod makes testing a bit more convenient. I'll clean this part of the code up a bit and add some tests to actually take advantage of that choice, haha.

examples/llm/tests/test_model.py
alextrott16: Added a combined prefix_lm/sequence_id test that should only pass if both of them are working as expected.

vchiley (approved):
lgtm

examples/llm/src/models/mosaic_gpt/mosaic_gpt.py
vchiley: I guess if eval ds passes `sequence_id` to fwd, then it is used right?

PR #268, opened 2023-03-31
Add bin packing collator wrapper in denoising

alextrott16:
Provides way to get concatenation and `"sequence_id"` support from raw text data in `denoising.py`.

Note: this PR also folds in a minor feature addition to the mixture-of-denoisers collator, which controls whether an EOS tag is appended to the end of the context (i.e. "prefix"). This feature is off by default.

### DIFF
--- a/examples/llm/src/data/denoising.py
+++ b/examples/llm/src/data/denoising.py
@@ -828,7 +828,7 @@ if __name__ == '__main__':
         'dataset': {
             'local': local,
             'remote': remote,
-            'split': 'val_small',
+            'split': 'val',  # 'val_small',
             'shuffle': False,
             'tokenizer_name': 'gpt2' if decoder_only else 't5-base',
             'max_seq_len': 2048 if decoder_only else 1024,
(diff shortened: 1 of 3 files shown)


examples/llm/src/data/denoising.py
```diff
@@ -143,6 +150,13 @@ def __init__(
         self._allow_pad_trimming = allow_pad_trimming
         self._seen_first_batch = False
 
+        if decoder_only_format:
+            if context_eos is None:
+                context_eos = False
+        else:
+            context_eos = True
+        self.context_eos = bool(context_eos)
+
```
vchiley, line 159:
```suggestion
        self.context_eos = bool(context_eos) if decoder_only_format else True

```

abhi-mosaic (approved):
Approving to unblock so we can deliver this in v0.0.4 release. Since this is an default-off feature for UL2 work it seems safe to include. In the future I think it could be useful to add tests for the binpacking, right now @alextrott16 I'm assuming you've visually inspected the packed batches and it matches the format you want.

alextrott16:
@abhi-mosaic Thanks! And, yeah, I've manually tested this and confirmed that it behaves as expected. The dataloader tests turn this feature on for both raw and pretokenized text, and it doesn't fail any of the asserts. I just pushed another commit that improves the printout for sanity checking and my sanity is confirmed intact :)

PR #271, opened 2023-04-02
updt throughput tables

vchiley:
This PR updates the [GPT throughput tables](https://github.com/vchiley/examples/tree/throughput_updt/examples/llm/throughput).

Note: since I branched from `main`, there are a few "extra" file which are already in relv04 branch. ignore


abhi-mosaic (approved):
Looks good, made small suggestion on the TODO messaging.

As always I wish we had more big cluster profiling results but we can add those as we get spare cycles.

PR #272, opened 2023-04-03
Make MosaicGPT MFU calculator more robust

abhi-mosaic:
Rather than cache the `self.num_fwd_flops`, we only cache `self.n_params` which is necessary to avoid FSDP sharding issues. And then we calculate the FLOPs dynamically on every forward pass for every batch (and we read the batch's `max_seq_len`.)

### DIFF
--- a/examples/llm/src/models/mosaic_gpt/mosaic_gpt.py
+++ b/examples/llm/src/models/mosaic_gpt/mosaic_gpt.py
@@ -498,8 +498,6 @@ class ComposerMosaicGPT(HuggingFaceModel):
             allow_embedding_resizing=True,
         )
 
-        self.num_fwd_flops = self._compute_num_fwd_flops()
-
         loss_fn_config = om_model_config.get('loss_fn', 'fused_crossentropy')
         if loss_fn_config == 'fused_crossentropy':
             try:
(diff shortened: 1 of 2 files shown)


examples/llm/src/models/mosaic_gpt/mosaic_gpt.py
```diff
@@ -542,20 +542,15 @@ def loss(self, outputs, batch):
...
-
     def flops_per_batch(self, batch):
         # Note: this computation does not take into account padding, and assumes
         # that the dataset has been constructed without padding. Additionally, we
         # assume the backward pass is approximately 2x the forward pass
-        return self.num_fwd_flops * 3 * batch['input_ids'].shape[0]
+
+        bs, msl = batch['input_ids'].shape[0:2]
+        params_flops_per_token = 2 * self.n_params
+        params_flops_per_seq = params_flops_per_token * msl
```
dakinggg, line 552: since we're doing this per batch now, and the tokenizer is stashed on the model, we might as well take padding into account now?

abhi-mosaic: If you feel confident in the implementation details then happy to try and edit that in... AFAIK, only the `flash` attention implementation does an unpadding step, right? So then this calculation is correct for torch/triton but potentially an overestimate for `flash`?

dakinggg: I think its fine to not do this as part of this pr/release and just jira it for now. but two things, 1) the current implementation is an overestimate (if using padding) for all right? because we don't want to count compute wasted on padding. I guess I'm not 100% sure if MFU is supposed to include compute on padding or not...but I would think not. if you get the "real" sequence length (different per item in batch), then the calculation becomes correct for all implementations. 2) I think the rough implementation is to, for each seq in batch, count the number of padding tokens, and subtract that from the max seq len to get the real seq len.

vchiley (commented):
making MFU based on `self.n_params` is a bad idea since this wont work for all model types (ie MoE)

`self.n_flops_per_token` could work? or `self.n_active_params`?

examples/llm/src/models/mosaic_gpt/mosaic_gpt.py
abhi-mosaic: I think we should do a bit of a design review on this one, so I'm gonna punt for now. Personally I try to think about MFU and "what tokens you ask the model to perform math on" as separate questions, so I would be ok with a batch of all pad tokens still being reported as 50% MFU aka the framework has computed the requested model FLOPs at a particular utilization. At the same time this batch of all padding would be like, a poor algorithmic choice of feeding batches to your model.

Basically let's do a review and push an update in v0.0.5!

PR #280, opened 2023-04-05
Improve bin packing efficiency

alextrott16:
Tweaks the bin packing collator to enable (substantially) more efficient packing -- which reduces both padding and data waste.


abhi-mosaic (approved):
LGTM

PR #282, opened 2023-04-05
make LlamaTokenizer work for tokenizer bakeoff

samhavens:
This adds some HF code that we can delete once we upgrade to a currently unreleased version that supports a certain type of sentencepiece tokenizer.

It also adds sentencepiece as an llm dependency, which is coming soon anyway.

Finally, it imports this code into common/convert_dataset.py and common/text_data.py to make it so AutoTokenizer.from_pretrained can be called with our new custom tokenizers and the Llama tokenizer

### DIFF
--- a/examples/llm/src/models/utils/make_llama_work_with_autoclass.py
+++ b/examples/llm/src/models/utils/make_llama_work_with_autoclass.py
@@ -81,11 +81,11 @@ class LlamaConfig(PretrainedConfig):
 class C4SentencePieceConfig(LlamaConfig):
 
     def __init__(self, *args, **kwargs):
+        kwargs['vocab_size'] = 65500
+        kwargs['pad_token_id'] = 3
+        kwargs['tie_word_embeddings'] = False
         super.__init__(
             *args,
-            vocab_size=65500,
-            pad_token_id=3,
-            tie_word_embeddings=False,
             **kwargs,
         )
 
@@ -224,9 +224,10 @@ class LlamaTokenizer(PreTrainedTokenizer):
         out_string += self.sp_model.decode(current_sub_tokens)
         return out_string
 
-    def save_vocabulary(self,
-                        save_directory,
-                        filename_prefix: Optional[str] = None) -> Tuple[str]:
+    def save_vocabulary(
+            self,
+            save_directory,
+            filename_prefix: Optional[str] = None) -> Optional[Tuple[str]]:
         """Save the vocabulary and special tokens file to a directory.
 
         Args:
--- a/examples/llm/src/models/utils/make_llama_work_with_autoclass.py
+++ b/examples/llm/src/models/utils/make_llama_work_with_autoclass.py
@@ -83,7 +83,6 @@ class C4SentencePieceConfig(LlamaConfig):
     def __init__(self, *args, **kwargs):
         kwargs['vocab_size'] = 65500
         kwargs['pad_token_id'] = 3
-        kwargs['tie_word_embeddings'] = False
         super.__init__(
             *args,
             **kwargs,
--- a/examples/llm/src/models/utils/make_llama_work_with_autoclass.py
+++ b/examples/llm/src/models/utils/make_llama_work_with_autoclass.py
@@ -78,17 +78,6 @@ class LlamaConfig(PretrainedConfig):
         )
 
 
-class C4SentencePieceConfig(LlamaConfig):
-
-    def __init__(self, *args, **kwargs):
-        kwargs['vocab_size'] = 65500
-        kwargs['pad_token_id'] = 3
-        super.__init__(
-            *args,
-            **kwargs,
-        )
-
-
 ##
 # tokenizer
 ##
@@ -319,8 +308,7 @@ class LlamaTokenizer(PreTrainedTokenizer):
 
 # as long as this line is invoked, you can call AutoTokenizer.from_pretrained on a
 # model that uses LlamaTokenizer
-AutoTokenizer.register(C4SentencePieceConfig,
-                       slow_tokenizer_class=LlamaTokenizer)
+AutoTokenizer.register(LlamaConfig, slow_tokenizer_class=LlamaTokenizer)
 
 
 def make_llama_work():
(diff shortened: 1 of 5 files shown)


examples/llm/src/models/utils/make_llama_work_with_autoclass.py
```diff
@@ -0,0 +1,329 @@
...
+    def __init__(
+        self,
+        vocab_file,
+        unk_token='<unk>',
+        bos_token='<s>',
+        eos_token='</s>',
+        pad_token=None,
+        sp_model_kwargs: Optional[Dict[str, Any]] = None,
+        add_bos_token=True,
+        add_eos_token=False,
```
samhavens, line 138: We may want to flip these @abhi-mosaic ; these are just the Llama defaults

examples/llm/src/models/utils/make_llama_work_with_autoclass.py
```diff
@@ -0,0 +1,329 @@
...
+        'hf-internal-testing/llama-tokenizer':
+            'https://huggingface.co/hf-internal-testing/llama-tokenizer/resolve/main/tokenizer_config.json',
+    },
+}
+PRETRAINED_POSITIONAL_EMBEDDINGS_SIZES = {
+    'hf-internal-testing/llama-tokenizer': 2048,
+}
+
+
+class LlamaTokenizer(PreTrainedTokenizer):
```
dakinggg, line 116: no fast?

examples/llm/src/models/utils/make_llama_work_with_autoclass.py
```diff
@@ -0,0 +1,329 @@
+# Copyright 2022 MosaicML Examples authors
```
dakinggg, line 1: I didn't really review this file, assuming it is copied

examples/llm/src/models/utils/make_llama_work_with_autoclass.py
samhavens: No, we need a newer version of `tokenizers` to do that, and we were worried about changing deps

examples/llm/src/models/utils/make_llama_work_with_autoclass.py
samhavens: there are ~10 lines at the end and `class C4SentencePieceConfig(LlamaConfig):` that are not copied

examples/llm/src/models/utils/make_llama_work_with_autoclass.py
```diff
@@ -0,0 +1,330 @@
...
+            **kwargs,
+        )
+
+
+class C4SentencePieceConfig(LlamaConfig):
+
+    def __init__(self, *args, **kwargs):
+        kwargs['vocab_size'] = 65500
+        kwargs['pad_token_id'] = 3
+        kwargs['tie_word_embeddings'] = False
```
dakinggg, line 86: whats up with this tie embeddings false thing? i assume it doesn't affect the tokenizer at all....but still wondering.

dakinggg (commented):
lgtm, will approve once we've resolved eos/bos

examples/llm/src/models/utils/make_llama_work_with_autoclass.py
samhavens: I meant for that to say True... I guess Llama didn't tie embeddings, and I was going to change that since we do. then I used the old value, so it both did nothing and was confusing

examples/llm/src/models/utils/make_llama_work_with_autoclass.py
```diff
@@ -0,0 +1,329 @@
...
+
+class C4SentencePieceConfig(LlamaConfig):
+
+    def __init__(self, *args, **kwargs):
+        kwargs['vocab_size'] = 65500
+        kwargs['pad_token_id'] = 3
+        super.__init__(
+            *args,
+            **kwargs,
+        )
```
abhi-mosaic, line 89: Why is this additional config necessary? When we build our custom tokenizer won't that just use a specific value of `vocab_size` and `pad_token_id`?

examples/llm/src/models/utils/make_llama_work_with_autoclass.py
```diff
@@ -0,0 +1,329 @@
...
+# as long as this line is invoked, you can call AutoTokenizer.from_pretrained on a
+# model that uses LlamaTokenizer
+AutoTokenizer.register(C4SentencePieceConfig,
+                       slow_tokenizer_class=LlamaTokenizer)
+
+
+def make_llama_work():
+    # it was actually just importing this class that makes it work
+    # this is so pyrite is happy
+    pass
```
abhi-mosaic, line 329: I'm a bit worried about this autoregistering behavior... are we hiding the lines that a customer would have to write in order to use our model? Also, do we have to add these lines to any of the `mosaic_gpt.py` or `hf_causal_lm.py` files?

abhi-mosaic (commented):
Couple questions, I'm not sure if I understand the full workflow yet of how we load / distribute our custom trained weights of the LLaMaTokenizer class

examples/llm/src/models/utils/make_llama_work_with_autoclass.py
codestar12: I think the goodnews is we won't have to do this for very long. We are only _just_ out ahead of a new transformers release, at which point we will have to do none of this (I think) and our custom tokenizers will just work.

jfrankle:
Are there continued blockers on merging this? It would be a big help to me for running experiments if this is checked in.

examples/llm/src/models/utils/make_llama_work_with_autoclass.py
samhavens: Resolution: BOS is fine

samhavens:
> Are there continued blockers on merging this? It would be a big help to me for running experiments if this is checked in.

Sorry, had to throw together a small seder and put kids to bed

jfrankle:
I think that's much much more important :)

jfrankle:
If you're enjoying the first night of Passover, go focus on that!

samhavens:
ok so this works, but there's a case where text_data.py will error with the sp tokenizers bc they dont set a pad token id, which apparently don't survive the HF conversion. still seems ok to merge

PR #283, opened 2023-04-06
Updates the export for onnx script to work with HF models

dakinggg:
- updates the onnx export script to work directly with HF models
- fixes some onnx incompatibilities in mosaicgpt
- adds an onnx export unit test

1b training on this pr: https://wandb.ai/mosaic-ml/daniel-debug/runs/3rnsjn09?workspace=user-danielking
pytest passes: `========================================================== 145 passed, 3 skipped, 31 xfailed, 352 warnings in 311.60s (0:05:11) ===========================================================
`


examples/llm/inference/convert_hf_to_onnx.py
```diff
@@ -0,0 +1,167 @@
+# Copyright 2022 MosaicML Examples authors
+# SPDX-License-Identifier: Apache-2.0
+
+"""Basic HuggingFace -> ONNX export script.
```
dskhudia, line 4: Could you update this to say it also supports exporting mosaicGPT checkpoints stored in HF compatible format?

examples/llm/inference/convert_hf_to_onnx.py
```diff
@@ -0,0 +1,167 @@
...
+
```
dskhudia, line 5: Add some example of invoking this script here like the ones we had in the previous script.

examples/llm/inference/convert_hf_to_onnx.py
dskhudia: Currently it sounds like it is for exporting HF models to onnx.

dskhudia (approved):

examples/llm/inference/convert_hf_to_onnx.py
dakinggg: it should work generally for HuggingFace models, but yes, I will add some language about MosaicGPT

examples/llm/inference/convert_hf_to_onnx.py
dakinggg: will do

dskhudia:
Would the onnx model work with kv caching? It's not returning past_key_values.

dakinggg:
No, I don't think this works with kv caching right now. I think HF has some support for doing this with `https://huggingface.co/docs/transformers/v4.27.2/en/main_classes/onnx#transformers.onnx.OnnxConfigWithPast`, but saw the point of this PR more as making sure that our model ops are able to be exported.

PR #285, opened 2023-04-06
enable user to pass in save_filename

vchiley:
using 
```save_filename: ep{epoch}-ba{batch}/rank{rank}.pt```
puts sharded ckpts in folder structure.


abhi-mosaic (approved):
LGTM!

PR #288, opened 2023-04-07
Add Monolithic Checkpoint Callback

abhi-mosaic:
Callback to on the side save monolithic checkpoints.
Manually tested
Usage:

```
callbacks:
    mono_ckpt_saver:
      save_folder: s3://mosaicml-internal-checkpoints-test/evan-test/checkpoints/ex-mono-test5/mono-{run_name}
      filename: ba{batch}/monlith.pt
      batch_interval: 2
```


vchiley:
checked I can restart from sharded ckpt and verified the mono ckpt can be loaded into an hf model and used for generating txt

PR #291, opened 2023-04-08
Fix generation callback interval

dakinggg:


examples/common/generate_callback.py
```diff
@@ -43,7 +45,14 @@ def init(self, state: State, logger: Logger):
                 if isinstance(destination, WandBLogger):
                     self.wandb_logger = destination
 
-    def eval_end(self, state: State, logger: Logger):
+    def batch_checkpoint(self, state: State, logger: Logger):
+        if (state.timestamp.batch.value % self.batch_log_interval) == 0:
+            self.generate(state, logger)
+
+    def fit_end(self, state: State, logger: Logger):
+        self.generate(state, logger)
```
vchiley, line 53:
```suggestion
        if (state.timestamp.batch.value % self.batch_log_interval) != 0:
            self.generate(state, logger)
```

prevents 2x generates at `fit_end`.

vchiley (approved):
lgtm

would adding a default batch_log_interval which only generates at the end of training make sense?

examples/common/generate_callback.py
vchiley: feel free to ignore if it doesn't make sense.

dakinggg: nah, makes sense

vchiley (approved):
lgtm

vchiley (approved):

PR #299, opened 2023-04-13
updt SchedGC to include cuda.empty_cache

vchiley:
add `torch.cuda.empty_cache` with gc.collect calls

### DIFF
--- a/examples/common/scheduled_gc_callback.py
+++ b/examples/common/scheduled_gc_callback.py
@@ -3,8 +3,17 @@
 
 import gc
 
+import torch
 from composer.core import Callback, State
 from composer.loggers import Logger
+
+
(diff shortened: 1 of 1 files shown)


examples/common/scheduled_gc_callback.py
```diff
@@ -3,10 +3,19 @@
 
 import gc
 
+import torch
 from composer.core import Callback, State
 from composer.loggers import Logger
 
 
+# from: https://github.com/BlackHC/toma/blob/10cfe70efaba59ea669c50c0060cfddef65d0b16/toma/torch_cuda_memory.py#L10
```
dakinggg, line 11: Idk if you really need to credit this, it's pretty common code I think

dakinggg (approved):

examples/common/scheduled_gc_callback.py
vchiley: ok

PR #301, opened 2023-04-13
Add example code to download huggingface checkpoint from OCI for generate 

sashaDoubov:
This PR shows a commented-out snippet of downloading hf checkpoint from OCI. The only caveat is that `bulk-download` downloads the entire prefix tree, so this adds a `mv` to be consistent with s3.


dakinggg (approved):
LGTM!

PR #302, opened 2023-04-13
add multi-query attn

vchiley:
add multi-query attn

test: added to `test_attn_impl`

### DIFF
--- a/examples/llm/src/models/layers/attention.py
+++ b/examples/llm/src/models/layers/attention.py
@@ -39,11 +39,13 @@ def scaled_multihead_dot_product_attention(
     dropout_p=0.0,
     training=False,
     needs_weights=False,
+    multiquery=False,
 ):
 
     q = rearrange(query, 'b s (h d) -> b h s d', h=n_heads)
-    k = rearrange(key, 'b s (h d) -> b h d s', h=n_heads)  # includes key.t()
-    v = rearrange(value, 'b s (h d) -> b h s d', h=n_heads)
+    k = rearrange(key, 'b s (h d) -> b h d s',
+                  h=1 if multiquery else n_heads)  # includes key.t()
+    v = rearrange(value, 'b s (h d) -> b h s d', h=1 if multiquery else n_heads)
 
     min_val = torch.finfo(q.dtype).min
 
@@ -122,6 +124,7 @@ def flash_attn_fn(
     dropout_p=0.0,
     training=False,
     needs_weights=False,
+    multiquery=False,
 ):
     try:
         from flash_attn import bert_padding  # type: ignore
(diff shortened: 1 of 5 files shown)


vchiley:
With multi-query attention, the key value projection matrix is smaller.
There are fewer Params / FLOPs in the model therefore the throughput is higher, but the MFU is pretty much the same
Screenshot 2023-04-17 at 10 12 54 PM
This is true for attn_impl \in {flash, triton, and torch}
Screenshot 2023-04-17 at 10 14 02 PM

Given there are fewer params, the networks final perf is degraded:
Screenshot 2023-04-17 at 5 00 19 PM
(granted the hparams are set for the multihead attn)


Generally the layer works. I wouldn't recommend but it might be useful in some inference contexts (BW or memory constrained inference settings)

examples/llm/src/models/layers/attention.py
```diff
@@ -347,6 +379,129 @@ def forward(self,
...
+    def __init__(
+        self,
+        d_model: int,
+        n_heads: int,
+        attn_impl: str = 'triton',
+        attn_clip_qkv: Optional[float] = None,
+        attn_qk_ln: bool = False,
+        softmax_scale: Optional[float] = None,
+        attn_pdrop: float = 0.0,
+        low_precision_layernorm: bool = False,
```
abhi-mosaic, line 398: this probably gets refactored after your norm PR goes in? Just heads up

examples/llm/src/models/layers/gpt_blocks.py
```diff
@@ -43,15 +44,17 @@ def __init__(self,
...
                  resid_pdrop: float = 0.0,
                  low_precision_layernorm: bool = False,
+                 multiquery_attention: bool = False,
                  device: Optional[str] = None,
                  **kwargs):
         del kwargs  # unused, just to capture any extra args from the config
         super().__init__()
 
         layernorm_class = LPLayerNorm if low_precision_layernorm else nn.LayerNorm
+        attn_class = MultiQueryAttention if multiquery_attention else MultiheadAttention
```
abhi-mosaic, line 54: Do you think we'll stop at 2? Or should we make `attn_class` a choice `str` out of [`multihead`, `multiquery`] the way we do `attn_impl`?

abhi-mosaic (approved):
Logic reads right to me. Leaning heavily on the WandB plots to say that convergence is happening (albeit a bit degraded). Cool to see the training throughput improvement. Would be very curious to try HF generate on a MQA model and see how much faster it is...

examples/llm/src/models/layers/gpt_blocks.py
vchiley: Yes, I wanted to do this as part of this PR, but this'll be a next step, wanted to merge this before the foundry migration.

examples/llm/src/models/layers/attention.py
vchiley: yeah merge conflicts make this one pretty clear...

PR #308, opened 2023-04-25
enable diff norm types

vchiley:
pass norm type and add rmsnorm as an option

issue: https://mosaicml.atlassian.net/browse/RESEARCH-572

### DIFF
--- a/examples/llm/src/models/layers/norm.py
+++ b/examples/llm/src/models/layers/norm.py
@@ -1,10 +1,9 @@
 # Copyright 2022 MosaicML Examples authors
 # SPDX-License-Identifier: Apache-2.0
 import torch
-from torch import nn
-
 from composer.algorithms.low_precision_layernorm.low_precision_layernorm import \
     LPLayerNorm
+from torch import nn
 
 
 def rms_norm(x, weight=None, eps=1e-5):
@@ -15,11 +14,20 @@ def rms_norm(x, weight=None, eps=1e-5):
 
 
 class RMSNorm(torch.nn.Module):
-    def __init__(self, normalized_shape, eps=1e-5, weight=True, dtype=None, device=None):
+
+    def __init__(
+        self,
+        normalized_shape,
+        eps=1e-5,
+        weight=True,
+        dtype=None,
+        device=None,
+    ):
         super().__init__()
         self.eps = eps
         if weight:
-            self.weight = nn.Parameter(torch.ones(normalized_shape, dtype=dtype, device=device))
+            self.weight = nn.Parameter(
+                torch.ones(normalized_shape, dtype=dtype, device=device))
         else:
             self.register_parameter('weight', None)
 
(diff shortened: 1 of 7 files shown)


examples/llm/src/models/layers/norm.py
```diff
@@ -0,0 +1,86 @@
+# Copyright 2022 MosaicML Examples authors
+# SPDX-License-Identifier: Apache-2.0
+import torch
+from composer.algorithms.low_precision_layernorm.low_precision_layernorm import \
```
dakinggg, line 4: can you copy lpln from composer into examples as part of this pr? it makes it easier to remove the composer dependency for hf release

examples/llm/src/models/layers/norm.py
```diff
@@ -0,0 +1,118 @@
...
+
+NORM_CLASS_REGISTRY = {
+    'layernorm': torch.nn.LayerNorm,
+    'ln': torch.nn.LayerNorm,  # alias
+    'low_precision_layernorm': LPLayerNorm,
+    'lplayernorm': LPLayerNorm,  # alias
+    'lpln': LPLayerNorm,  # alias
+    'rmsorm': RMSNorm,
+    'low_precision_rmsorm': LPRMSNorm,
+    'lprmsorm': LPRMSNorm,  # alias
```
abhi-mosaic, line 117: will having aliases make it confusing for people if they save two models with configs who differ by an alias, but are really using the same norm_type?

dakinggg (commented):
Add a unit test for constructing the diff norm types please :)

examples/llm/src/models/layers/norm.py
```diff
@@ -0,0 +1,86 @@
...
+
+NORM_CLASS_REGISTRY = {
+    'layernorm': nn.LayerNorm,
+    'ln': nn.LayerNorm,  # alias
+    'low_precision_layernorm': LPLayerNorm,
+    'lplayernorm': LPLayerNorm,  # alias
+    'lpln': LPLayerNorm,  # alias
+    'rmsorm': RMSNorm,
+    'low_precision_rmsorm': LPRMSNorm,
+    'lprmsorm': LPRMSNorm,  # alias
```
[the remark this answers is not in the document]
vchiley: per abhi's suggestion, going to rm alias versions.

examples/llm/src/models/mosaic_gpt/configuration_mosaic_gpt.py
```diff
@@ -44,7 +44,7 @@ def __init__(
         fan_mode: str = 'fan_in',
         init_nonlinearity: str = 'relu',
         embedding_fraction: float = 1.0,
-        low_precision_layernorm: bool = True,
+        norm_type: str = 'low_precision_layernorm',
```
dakinggg, line 47: should we accept the old arg and just map it to the correct norm type for backwards compat reasons?

vchiley: what you're describing is just a config updt; I wouldn't worry about it.


Changing this
```
        self.ln_1 = layernorm_class(d_model, device=device)
```
to this
```
        self.norm_1 = norm_class(d_model, device=device)
```
is the bigger bwd compat issue.
(param names from old models will be diff; cant just load them)

dakinggg: that is fair

dakinggg (approved):
lgtm

PR #317, opened 2023-05-02
Remove deprecated code

mvpatel2000:
Remove deprecated code related to grad_accum and old precision names

Repoint LLM code


examples/common/config_utils.py
```diff
@@ -9,39 +9,8 @@
 from omegaconf import OmegaConf as om
 
 
-def calculate_batch_size_info(global_batch_size: int,
```
dakinggg, line 12: the error message in this function still seems useful?

mvpatel2000:
Oops... still need to be bump... it breaks a few tests...

.github/workflows/lint.yaml
```diff
@@ -18,6 +18,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 20
     strategy:
+      fail-fast: false  # TODO: Delete
```
dakinggg, line 21: delete?

.github/workflows/test.yaml
```diff
@@ -18,6 +18,7 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 20
     strategy:
+      fail-fast: false  # TODO: Delete
```
dakinggg, line 21: delete?

examples/bert/tests/test_main.py
```diff
@@ -22,9 +22,9 @@ def test_trainer(model_name: str, seed: int):
 
     with SynthTextDirectory() as tmp_datadir:
         config.train_loader.dataset.remote = tmp_datadir
-        config.train_loader.dataset.local = tmp_datadir
+        config.train_loader.local = os.path.join(tmp_datadir, 'local2')
```
dakinggg, line 25: i don't think removing the `dataset` part of the prefix here is correct?

examples/bert/tests/test_main.py
```diff
@@ -22,9 +22,9 @@ def test_trainer(model_name: str, seed: int):
...
         config.eval_loader.dataset.remote = tmp_datadir
-        config.eval_loader.dataset.local = tmp_datadir
+        config.eval_loader.local = os.path.join(tmp_datadir, 'local2')
```
dakinggg, line 27: same

examples/bert/tests/test_main.py
```diff
@@ -40,6 +40,8 @@ def test_trainer(model_name: str, seed: int):
         # Check that the checkpoint was loaded by comparing model weights (with no weight changes)
         config.load_path = chkpt_path
         config.seed += 10  # change seed
+        config.train_loader.local = os.path.join(tmp_datadir, 'local2')
```
dakinggg, line 43: same

requirements.txt
```diff
@@ -2,7 +2,7 @@
 # Each example subdir has its own `requirements.txt` with extra deps.
 # E.g. to install LLM folder deps, use `pip install -e .[llm]`
 
-mosaicml>=0.13.4,<0.14
+mosaicml>=0.13.4,<0.15
```
dakinggg, line 5: should this be bumped to 0.14 min?

dakinggg (approved):

requirements.txt
mvpatel2000: llm is pinned to lower version still (not bothering to upgrade since we'll remove) so this is looser

PR #318, opened 2023-05-03
Polish inference example README

dblalock:
Reference the correct config file name, mention installing dependencies, be clear about what requirements custom models have to meet.


margaretqian (approved):
thank you!

PR #323, opened 2023-05-10
Adding more inference model handlers

margaretqian:
Moved over from https://github.com/mosaicml/inference-model-registry/pull/1

PR #324, opened 2023-05-12
Update inference model handlers with suggested comments

margaretqian:
Comments from https://github.com/mosaicml/examples/pull/323

### DIFF
--- a/.github/workflows/lint.yaml
+++ b/.github/workflows/lint.yaml
@@ -32,6 +32,7 @@ jobs:
         - nemo
         - gpt_neox
         - trlx
+        - inference_model_handlers
     steps:
     - uses: actions/checkout@v3
     - uses: actions/setup-python@v4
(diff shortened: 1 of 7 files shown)


margaretqian:
Answering some questions from the previous PR:

> What is difference between download_parameters and downloader?
@mvpatel2000  the downloader is the path to the download function and download_parameters are the inputs that get passed in

> why does the hf model name need to be specified in multiple places?
@dakinggg the first reference to the hf model name is for downloading the checkpoint files and the second in model_parameters tells the handler which instructor model we want to load (since both the large and xl model can share the same handler)

dakinggg (commented):
One comment that I don't want to get lost (although doesn't necessarily need to happen in this PR), `generate_kwargs['eos_token_id'] = self.tokenizer.eos_token_id`, is not ideal for a couple reasons. As a user, I want to be able to specify arbitrary stop tokens, and the `eos_token_id` arg accepts a list (https://huggingface.co/docs/transformers/v4.29.1/en/main_classes/text_generation#transformers.GenerationConfig.eos_token_id). For this list, I don't want to have to specify the token ids, but rather the strings themselves, because I don't want to have to instantiate the tokenizer client side. So, I'd like to specify a list of strings like `['\n', '#', '.']`, and have the server convert those to ids (it will need to verify that each string converts to a single id and error if it does not), and then additionally add the `self.tokenizer.eos_token_id` (maybe if i specify `None` it doesn't add the eos token id either.

Will approve after CI is passing

PR #333, opened 2023-05-24
Begin restructuring

dakinggg:
This PR restructures the examples repo to prepare for the further changes and new examples.
It
1. creates the new folders
2. updates the top level readme
3. remove the pip installable nature of the whole repo, and instead uses requirements.txt in each subfolder
4. updates the lint/test scripts so they still pass

There are many improvements to be made following this PR, but this PR should be "complete" in the sense that everything that already existed in examples repo is still runnable (except llm, which has been officially deleted)

Note: mcli yamls are going to remain pinned to v0.0.4 for now, so they will not change in this PR.

TODOs:
- [x] grep for `examples.` and fix
- [x] bert readme runthrough
- [x] deeplab readme runthrough
- [x] imagenet readme runthrough
- [x] cifar readme runthrough
- [x] diffusion readme runthrough
- [x] dreambooth readme runthrough
- [x] update top level readme
- [x] lint passing
- [x] unittests passing
- [x] inference deployment checked
- [x] fix all pip install commands throughout codebase


scripts/test_subdirectory.sh
```diff
@@ -19,8 +19,21 @@ source "$ENV_NAME/bin/activate"
...
+original_dir=$(pwd)
+cd examples/$1
+if [ -f requirements-cpu.txt ]; then
+    pip install -r requirements-cpu.txt
+elif [ -f requirements.txt ]; then
+    pip install -r requirements.txt
+else
+    echo "No requirements-cpu.txt or requirements.txt found in directory examples/$1"
+fi
+cd $original_dir
```
mvpatel2000, line 36: We really should just rework this... but fine with follow-on PR

examples/benchmarks/bert/__init__.py
```diff
@@ -0,0 +1,73 @@
...
+    if torch.cuda.is_available():
+        from src.flash_attn_triton import \
+            flash_attn_func as flash_attn_func_bert # type: ignore
+        from src.flash_attn_triton import \
+            flash_attn_qkvpacked_func as flash_attn_qkvpacked_func_bert # type: ignore
+    from src.hf_bert import create_hf_bert_classification, create_hf_bert_mlm
+    from src.mosaic_bert import (create_mosaic_bert_classification,
+                                 create_mosaic_bert_mlm)
+except ImportError as e:
+    try:
```
mvpatel2000, line 32: is there a reason we're not removing this now?

examples/benchmarks/bert/glue.py
```diff
@@ -13,46 +13,105 @@
 from typing import Any, Dict, List, Optional, Sequence, Set, Tuple
 from urllib.parse import urlparse
 
+sys.path.append(os.path.dirname(os.path.realpath(__file__)))
```
mvpatel2000, line 16: can we put a comment where we do this (like what I did for tests with scripts in LLM foundry?)

examples/benchmarks/bert/__init__.py
dakinggg: ah i missed that one. still need the message i think but the message should change to reflect the new install method.

scripts/test_subdirectory.sh
dakinggg: i think this is actually a fine way to do it? but yeah, either way, not in this pr.

examples/benchmarks/bert/glue.py
dakinggg: will do

dakinggg: added comments, and also fixed a few of them

.github/workflows/test.yaml
```diff
@@ -22,25 +22,22 @@ jobs:
...
-          subdir: resnet_imagenet
+          subdir: benchmarks/resnet_imagenet
         - name: stable-diffusion-pytest
           container: mosaicml/pytorch:1.13.1_cpu-python3.10-ubuntu20.04
-          subdir: stable_diffusion
+          subdir: end-to-end-examples/stable_diffusion
         - name: stable-diffusion-dreambooth-pytest
           container: mosaicml/pytorch:1.13.1_cpu-python3.10-ubuntu20.04
-          subdir: stable_diffusion_dreambooth
+          subdir: end-to-end-examples/stable_diffusion_dreambooth
```
mvpatel2000, line 40: can we change to pytorch_vision:latest_cpu so we dont have to keep updating

dakinggg: that would mean that examples repo doesn't get to decide when it upgrades to torch2 images, which im not sure im a fan of. wdyt?

mvpatel2000: ok fair

examples/__init__.py
```diff
@@ -1,4 +1,2 @@
 # Copyright 2022 MosaicML Examples authors
 # SPDX-License-Identifier: Apache-2.0
-
-__version__ = '0.0.4'
```
mvpatel2000, line 4: Should we leave versioning even if its not pip installable? We will still have "release" tags...

examples/benchmarks/resnet_imagenet/main.py
```diff
@@ -7,6 +7,11 @@
 import sys
 from typing import Dict
 
+# Add folder root to path to allow us to use relative imports regardless of what directory the script is run from
+sys.path.append(os.path.dirname(os.path.realpath(__file__)))
+
+import data as data_module
+import model as model_module
```
mvpatel2000, line 14: why this instead of importing the necessary functions directly

examples/benchmarks/resnet_imagenet/main.py
```diff
@@ -94,9 +106,10 @@ def main(config):
 
     # Instantiate torchvision ResNet model
     print('Building Composer model')
-    composer_model = build_composer_resnet(model_name=config.model.name,
-                                           loss_name=config.model.loss_name,
-                                           num_classes=config.model.num_classes)
+    composer_model = model_module.build_composer_resnet(
+        model_name=config.model.name,
+        loss_name=config.model.loss_name,
+        num_classes=config.model.num_classes)
```
mvpatel2000, line 112:
```suggestion
        num_classes=config.model.num_classes,
      )
```

examples/benchmarks/resnet_imagenet/tests/test_data.py
```diff
@@ -1,14 +1,19 @@
...
 import pytest
 import torch
 
-from examples.resnet_imagenet.data import (build_imagenet_dataspec,
-                                           check_dataloader)
-from examples.resnet_imagenet.tests.utils import SynthClassificationDirectory
+# Add tests folder root to path to allow us to use relative imports regardless of what directory the script is run from
+sys.path.append(os.path.dirname(os.path.abspath(__file__)))
+# Add folder root to path to allow us to use relative imports regardless of what directory the script is run from
+sys.path.append(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
```
mvpatel2000, line 13: Why twice?

mvpatel2000 (approved):

examples/benchmarks/resnet_imagenet/tests/test_data.py
dakinggg: they're different! one is the tests folder, and one is the source folder

examples/benchmarks/resnet_imagenet/main.py
dakinggg: idk, seemed safer. not really attached to anything in particular, but would also rather not change it at this point

examples/__init__.py
dakinggg: I think deleting versioning is reasonable, but can keep release tags? Like, we definitely have to keep the old releases. Not sure we need to do new releases, except for maybe one more explaining the new structure? wdyt?

PR #336, opened 2023-05-25
MPT Handler Predict Fix

RR4787:
Predict was out of date with new server Forward method which sends a BatchedRequest with potentially multiple input dicts.
Predict now generates output for each input dict and returns list of outputs as expected by Forward. Predict was also returning output as bytes but proto Structs don't take bytes as a valid value so now predict returns output as string.


margaretqian (commented):
sorry getting to this late -- @Glabred might've already fixed this? https://github.com/mosaicml/examples/blob/main/examples/inference-deployments/mpt/mpt_7b_handler.py#L73

PR #338, opened 2023-05-28
update streaming

tbarton16:
Update the streaming library version. Tested here `mcli logs semdedupbaseline-oN3r4G  `                                             
`


dakinggg:
What is the motivation for updating the min pin here?

tbarton16:
> What is the motivation for updating the min pin here?

There is a fatal error if you have `shuffle_algo: py1b`

tbarton16:
Closing, moving change to llm-foundry

PR #341, opened 2023-06-05
Update inference models to work with batching and go server

RR4787:
updates yamls and model handlers to handle go server/batching. Leaving diffusion since its broken and needs to be fixed. 
Change attn_cnfg['attn_impl'] in mpt7b handler from 'triton' to 'torch' for the time being while dependency issues are sorted out in the go server

Manually tested deploying the models with go server and using a script to test batching requests.


dakinggg:
Could you please include some evidence that these work in the PR description? Ideally we would have tests that at least try to start the server in CI, but understand if not top priority right now, and would like some manual tests in the PR description at least.

examples/inference-deployments/mpt/mpt_7b_storywriter.yaml
```diff
@@ -4,8 +4,10 @@ gpu_type: a100_40gb
 image: mosaicml/inference:latest
 replicas: 1
 command: |
-  cd examples
-  PYTHONPATH=.. uvicorn serve:app --host 0.0.0.0 --port 8080
+  python3 inference_go/downloader.py &&
+  export PATH="$PATH:/usr/local/go/bin" &&
+  cd examples &&
+  PYTHONPATH=. go run ../inference_go/inference_server/cmd/main.go
```
margaretqian, line 10: we shouldn't need this anymore after switching to the go server

examples/inference-deployments/mpt/mpt_7b_handler.py
```diff
@@ -63,10 +63,33 @@ def _parse_inputs(self, inputs: Dict[str, Any]):
...
-        outputs = self.generator(generate_input, **generate_kwargs)
-        return [output[0]['generated_text'] for output in outputs]
+    def _extract_output(self, outputs: List[Any]):
+        output_list = []
+        for output in outputs:
+            output_bytes = output[0]['generated_text']
+            output_list.append(output_bytes)
+        return output_list
+
+    def predict(self, input_dicts: List[Dict[str, Any]]):
```
margaretqian, line 73: we should have this match the hf_model handler (https://github.com/mosaicml/mcloud/blob/dev/inference_go/model_server/hf_model.py#L146-L168)

margaretqian (changes requested):

examples/inference-deployments/mpt/mpt_7b_handler.py
RR4787: done

margaretqian (approved):

PR #342, opened 2023-06-07
Update example yaml to work without git secrets

margaretqian:


examples/inference-deployments/instructor/instructor_xl.yaml
```diff
@@ -8,8 +8,8 @@ command: |
   PYTHONPATH=.. uvicorn serve:app --host 0.0.0.0 --port 8080
 integrations:
 - integration_type: git_repo
-  git_repo: dakinggg/examples
-  git_branch: restructure
+  git_repo: mosaicml/examples
```
margaretqian, line 11: @dakinggg not sure if you intended to keep your branch here?

dakinggg: I did not 😬

dakinggg (approved):

PR #346, opened 2023-06-08
Update MPT FT handler to work with go server

asfandyarq:
Main changes:
- Update `predict` to work with new batched input format (adapted from #341).
- Better kwarg type conversion, to handle issues with json passed into this
- FT checkpoint reuse and more logging (helps with debugging)
- Explicitly import `FasterTransformer.examples` to avoid confusion with `mosaicml/examples`

Tested with latest inference images.

### DIFF
--- a/examples/inference-deployments/mpt/mpt_7b_ft_handler.py
+++ b/examples/inference-deployments/mpt/mpt_7b_ft_handler.py
@@ -5,13 +5,13 @@ import argparse
 import configparser
 import copy
 import os
-from typing import Any, Dict
+from typing import Any, Dict, List, Tuple
 
 import torch
 from torch.nn.utils.rnn import pad_sequence
 from transformers import AutoTokenizer
 
-from examples.pytorch.gpt.utils.parallel_gpt import ParallelGPT  # type: ignore
+from FasterTransformer.examples.pytorch.gpt.utils.parallel_gpt import ParallelGPT  # type: ignore
 
 from scripts.inference.convert_hf_mpt_to_ft import convert_mpt_to_ft  # isort: skip # type: ignore
 
@@ -22,7 +22,7 @@ class MPTFTModelHandler:
 
     DEFAULT_GENERATE_KWARGS = {
         # Output sequence length to generate.
-        'output_len': 256,
+        'max_length': 256,
         # Beam width for beam search
         'beam_width': 1,
         # top k candidate number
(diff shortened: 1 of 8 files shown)


examples/inference-deployments/mpt/mpt_7b_ft_handler.py
```diff
@@ -124,30 +131,50 @@ def __init__(self,
...
+
+    def _convert_kwargs(self, generate_inputs: List[str],
+                        generate_kwargs: Dict[str, Any]):
+        """Converts generate_kwargs into required torch types."""
+        batch_size = len(generate_inputs)
+
+        # Allow 'max_length' to be an alias for 'output_len'. Makes it less
+        # likely clients break when we swap in the FT handler.
+        if 'max_length' in generate_kwargs:
+            generate_kwargs['output_len'] = generate_kwargs['max_length']
```
dskhudia, line 171: wrap this by int() too.

dskhudia (approved):

dskhudia:
Looks great. Thank you for the improvements. Feel free to merge once CI is green.

examples/inference-deployments/mpt/mpt_7b_ft_handler.py
asfandyarq: that's redundant, right? since we'll always convert `output_len` on line 175.

dskhudia: True. Missed that.

examples/inference-deployments/mpt/mpt_7b_ft_handler.py
```diff
@@ -65,15 +65,21 @@ def __init__(self,
...
+            # Datatype of weights in the HF checkpoint
+            weight_data_type = 'fp32'
+            convert_mpt_to_ft(self.model_name, LOCAL_CHECKPOINT_PATH, gpus,
+                              weight_data_type, False)
+            if not os.path.isfile(ckpt_config_path):
+                raise RuntimeError('Failed to create FT checkpoint')
+        else:
+            print(f'Reusing existing FT checkpoint at {model_path}')
+        ckpt_config = configparser.ConfigParser()
+        ckpt_config.read(ckpt_config_path)
```
margaretqian, line 82: I think it might be cleaner to move this into a custom download function (like `download_convert_mpt_ft` and then set the `downloader` path in the input yaml to `examples.inference-deployments....download_convert_mpt_ft` -- this gets run first thing when the pod spins up, so we can fail early if there's any issue with the checkpoint conversion

examples/inference-deployments/mpt/mpt_7b_ft_handler.py
```diff
@@ -106,6 +112,7 @@ def __init__(self,
 
         self.end_id = end_id
 
+        print('Initializing FasterTransformer')
```
margaretqian, line 115: are these for debugging or do you intend to leave these in?

margaretqian (approved):
awesome!

examples/inference-deployments/mpt/mpt_7b_ft_handler.py
asfandyarq: I was going to leave these in. the steps between these prints sometimes take minutes, so it's useful to see these if you pull the deployment logs while waiting for the server to become ready.

examples/inference-deployments/mpt/mpt_7b_ft_handler.py
asfandyarq: yeah that sounds like a good idea. will use a follow-up PR for that.

PR #349, opened 2023-06-09
Modify model handlers to take in new request structure

Glabred:
The new list of inputs passed to the handler are of the following structure dependent on this PR https://github.com/mosaicml/mcloud/pull/1589:
```
[
  {
    "input": <str/int/whatever>,
    "parameters": Dict
  },
  {
    "input": <str/int/whatever>,
    "parameters": Dict
  },
]
```

Also in this PR: renamed `inputs` to `model_requests`

handlers tested with different parameters and work:
`mpt_7b_handler.py`
not tested:
`mpt_7b_ft_handler.py`
can't test:
`instuctor_handler.py`

### DIFF
--- a/examples/inference-deployments/instructor/instructor_handler.py
+++ b/examples/inference-deployments/instructor/instructor_handler.py
@@ -1,6 +1,6 @@
 # Copyright 2022 MosaicML Examples authors
 # SPDX-License-Identifier: Apache-2.0
-from typing import Any, Dict, List
+from typing import Dict, List
 
 import torch
 from InstructorEmbedding import INSTRUCTOR
@@ -18,7 +18,7 @@ class HFInstructorHandler():
         self.model = INSTRUCTOR(self.model_name)
         self.model.to(self.device)
 
-    def predict(self, model_requests: List[Dict[str, Any]]):
+    def predict(self, model_requests: List[Dict]):
         """Runs a forward pass with the given inputs.
 
         Model request format: {"input": ["<instruction>", "<sentence>"]}
--- a/examples/inference-deployments/instructor/instructor_handler.py
+++ b/examples/inference-deployments/instructor/instructor_handler.py
@@ -27,7 +27,7 @@ class HFInstructorHandler():
         for req in model_requests:
             if 'input' not in req:
                 raise KeyError('input key not in model_requests')
-            input_list.append(input['input'])
+            input_list.append(req['input'])
 
         embeddings = self.model.encode(input_list)
         return embeddings.tolist()
--- a/examples/inference-deployments/instructor/instructor_handler.py
+++ b/examples/inference-deployments/instructor/instructor_handler.py
@@ -18,16 +18,16 @@ class HFInstructorHandler():
         self.model = INSTRUCTOR(self.model_name)
         self.model.to(self.device)
 
-    def predict(self, inputs: List[Dict[str, Any]]):
+    def predict(self, model_requests: List[Dict[str, Any]]):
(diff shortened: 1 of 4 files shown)


examples/inference-deployments/mpt/mpt_7b_ft_handler.py
```diff
@@ -8,13 +8,14 @@
 from typing import Any, Dict, List, Tuple
 
 import torch
+# isort: skip # type: ignore
+from FasterTransformer.examples.pytorch.gpt.utils.parallel_gpt import \
+    ParallelGPT
+# isort: skip # type: ignore
```
asfandyarq, line 14: nit: since we're sorting these imports now, remove the `isort: skip` part?

examples/inference-deployments/mpt/mpt_7b_ft_handler.py
```diff
@@ -140,25 +142,21 @@ def __init__(self,
...
-            self, inputs: Dict[str, Any]) -> Tuple[List[str], Dict[str, Any]]:
-        """Splits request into input strings and kwargs."""
-        if self.INPUT_STRINGS_KEY not in inputs or not isinstance(
-                inputs[self.INPUT_STRINGS_KEY], list):
+    def _parse_model_request(
+            self, model_request: Dict[str, Any]) -> Tuple[str, Dict[str, Any]]:
+        if self.INPUT_KEY not in model_request:
             raise RuntimeError(
-                'Input strings must be provided as a list to generate call')
+                f'{self.INPUT_KEY} must be provided to generate call')
```
asfandyarq, line 149: nit: add quotes so it's more obvious this is a literal key: `f'"{self.INPUT_KEY}" ...'`.

examples/inference-deployments/instructor/instructor_handler.py
```diff
@@ -18,16 +18,16 @@ def __init__(self, model_name: str):
         self.model = INSTRUCTOR(self.model_name)
         self.model.to(self.device)
 
-    def predict(self, inputs: List[Dict[str, Any]]):
+    def predict(self, model_requests: List[Dict[str, Any]]):
```
asfandyarq, line 21: don't need to change anything, but: now that we're using a nested dict here, might be worth simplifying these annotations (`Dict` instead of `Dict[str, Any]`).

examples/inference-deployments/mpt/mpt_7b_ft_handler.py
```diff
@@ -219,8 +216,9 @@ def _parse_inputs(
 
         return generate_inputs, generate_kwargs
 
-    def predict(self, input_dicts: List[Dict[str, Any]]) -> List[str]:
-        generate_inputs, generate_kwargs = self._parse_inputs(input_dicts)
+    def predict(self, model_requests: List[Dict[str, Any]]) -> List[str]:
```
asfandyarq, line 219: the new request structure matches the [HF schema now](https://huggingface.github.io/text-generation-inference/#/Text%20Generation%20Inference/generate), right? worth documenting?

asfandyarq (approved):
LGTM (with minor comments)

examples/inference-deployments/mpt/mpt_7b_ft_handler.py
Glabred: This was completely by accident! We had no idea. I can see a subtle difference where we must take in a list of inputs while HF takes in just one input as a string. This is because we need to be compatible for other types of deployments not just inference for NLP.

PR #361, opened 2023-06-15
Fix mpt 7b handler

margaretqian:
Model needs to run with autocast now


dakinggg (approved):
lets make this pr just the handler fix, and any yamls updates in a follow up pr

PR #363, opened 2023-06-15
Fix for multi-gpu inference

dskhudia:
Fixes multi-gpu inference hang by adding torch.manual_seed(0)

### DIFF
--- a/examples/inference-deployments/mpt/mpt_7b_ft_handler.py
+++ b/examples/inference-deployments/mpt/mpt_7b_ft_handler.py
@@ -12,7 +12,9 @@ from urllib.parse import urlparse
 import boto3
 import botocore
 import torch
+import torch.distributed as dist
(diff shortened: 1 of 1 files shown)


Glabred (approved):

examples/inference-deployments/mpt/mpt_7b_ft_handler.py
```diff
@@ -171,6 +178,8 @@ def __init__(self,
 
         self.end_id = end_id
 
+        comm.initialize_model_parallel(tensor_para_size, pipeline_para_size)
```
margaretqian, line 181: do we need this?

margaretqian: we call `dist.init_process_group(backend=dist.Backend.MPI)` in the launcher tho -- won't this conflict?

asfandyarq (approved):

PR #375, opened 2023-06-20
update mpt7b instruct ft commit

RR4787:
Update yaml commit and pin the HF model commit


examples/inference-deployments/mpt/mpt_7b_ft_handler.py
```diff
@@ -22,6 +22,7 @@
 
 LOCAL_CHECKPOINT_DIR = '/tmp/mpt'
 LOCAL_MODEL_PATH = os.path.join(LOCAL_CHECKPOINT_DIR, 'local_model')
+HF_COMMIT = 'bbe7a55d70215e16c00c1825805b81e4badb57d7'
```
margaretqian, line 25: this should be passed in from the input yaml

margaretqian (commented):
the merge conflict is probably due to the file name changing

PR #382, opened 2023-06-21
[Do not merge] Hosted handler for MPT

ajaysaini725:
Opening this for review but do not merge


ajaysaini725:
Opening this for review but do not merge

examples/inference-deployments/mpt/mpt_hosted_handler.py
```diff
@@ -100,7 +100,7 @@ def download_convert(s3_path: Optional[str] = None,
 class MPTFTHostedModelHandler:
     # This is what the user request will contain
     INPUT_GENERATE_KWARGS = {
-        'max_new_tokens': 256,
+        'max_tokens': 256,
```
dskhudia, line 103: Did we end up on this? It is kind of confusing because it doesn't say whether it's generated or prompt + generated.

dakinggg: imo it should be `max_new_tokens`, `max_tokens` is not very helpful, because your prompt can be longer than 256 and then you just get a 500 from the server.

PR #397, opened 2023-06-22
Fix mpt 7b yamls

sakshigupta93:
We added commit hashes so that we don't need to change existing 7b yamls but then we changed the handler name for them. These are currently broken because the deployment keeps searching for mpt_handler and can't find it, because the commit hash has mpt_7b_handler and not mpt_handler. These changes were made in https://github.com/mosaicml/examples/pull/390


ajaysaini725 (approved):
🚀

PR #398, opened 2023-06-26
Add secrets check as part of pre-commit

karan6181:
# What does this PR do?
- Add secrets check as part of pre-commit.
- The `pre-commit run` command would fail if there is a verified secret leak. 

# What issue(s) does this change relate to?
CO-2201


bmosaicml:
LGTM!

bmosaicml (approved):

PR #399, opened 2023-06-27
Rename MOSAICML_API_TOKEN to MOSAICML_API_KEY

ajaysaini725:


dakinggg:
This is the env var name that I used in the langchain integration. I'm not sure why I named it that way, probably just a mistake, but this change requires a change to the langchain integration.

PR #411, opened 2023-07-06
Prototype outline

ajaysaini725:


examples/end-to-end-examples/finetuning/finetune_small_dataset.py
```diff
@@ -0,0 +1,52 @@
...
+  MPT_30B = 'mpt-30b'
+
+
+def main(model_name: str, train_data_path: str, save_folder: str):
+    # Load the finetuning config
+    config: RunConfig = None
+    if model_name == Models.MPT_7B.value:
+        # TODO: Copy / move this YAML to the same folder as this python script
+        # Note that we use this one because it loads a model with pretrained: true
+        config = RunConfig.from_file('../sec_10k_qa/mcli-yamls/03_finetune_on_10ks.yaml')
```
dakinggg, line 25: if you want to start with ift, this will need to be 04, but i would recommend defining your own config python dictionary, rather than importing from a different example.

dakinggg: that way this example can have a "base" config dictionary, and then we can fill in the details (model name, save folder, etc)

PR #425, opened 2023-07-27
Llama2 handler

margaretqian:


examples/inference-deployments/llama2/llama2_handler.py
```diff
@@ -0,0 +1,129 @@
...
+        'max_length': 256,
+        'use_cache': True,
+        'do_sample': True,
+        'top_p': 0.95,
+        'temperature': 0.8,
+    }
+
+    INPUT_KEY = 'input'
+    PARAMETERS_KEY = 'parameters'
+    MODEL_DTYPE = 'fp16'
```
asfandyarq, line 27: Unused?

examples/inference-deployments/llama2/llama2_handler.py
```diff
@@ -0,0 +1,129 @@
...
+
+    def __init__(
+        self,
+        model_name_or_path: str,
+    ):
```
asfandyarq, line 32: Could we add a `max_output_tokens` parameter here, or in the go server? We'd then clip user `max_length` values in requests to this. 

This is a slow model and we'll probably need to limit it to globally to 128 tokens. Otherwise someone could send a super long request and tie up the entire server for >30 sec.

asfandyarq (approved):

PR #428, opened 2023-08-01
Refine llama2 handler

asfandyarq:
Changes:
- Add `torch.inference_mode` decorators.
- Instead of using a `pipeline`, manually implement `tokenize -> generate`. The pipeline was serializing batches, giving us linearly increasing inference latency for batched requests.
- Additional logic around `max_new_tokens` (global cap, translation from `max_length`).


asfandyarq:
No longer relevant because we dropped deepspeed.

PR #432, opened 2023-08-28
Fix no code image version

margaretqian:


waiwuc (approved):

examples/inference-deployments/mpt/mpt_7b.yaml
```diff
@@ -2,7 +2,7 @@ name: mpt-7b-simple
 compute:
   gpus: 1
   instance: oci.vm.gpu.a10.1
-image: mosaicml/inference:0.1.16
+image: mosaicml/inference:0.1.36
```
tianshuc, line 5: Why not update the image version for all example configs?

margaretqian: the others aren't broken with this version, so we don't urgently need to update them -- we can update those when I do a pass to clean up examples that we're planning on deprecating

